PR #1, opened 2016-07-18
Run administrative tools

bumbadadabum:
Why bother with indenting and optimizing file size when you have tools that do that for you with a single command?


Dugy:
Thank you. I didn't even know of optipng.

PR #7, opened 2016-12-07
address diminishing money bug?

MPagel:
cash seems to backslide between turns, possibly when villages are taken or lost? Hopefully these patches will fix this for four scenarios. Don't have time right now to edit all scenarios with the text "no matter what it shows". untested

### DIFF
--- a/scenarios6/13_Phoenix.cfg
+++ b/scenarios6/13_Phoenix.cfg
@@ -97,6 +97,7 @@
         team_name=Good
         user_team_name=_"Good"
         village_gold=2
+        village_support=1000
         shroud=yes
     [/side]
     [side]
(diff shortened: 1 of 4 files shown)


Dugy:
Thank you. I think that the village_support option wasn't available at the time the campaign was coded.

MPagel:
there were reports of it not actually working in the early-going. Documentation doesn't seem to reveal if there's an upper limit for the value.  Again, this is untested.  Also, I'm not sure if you saw the top comment line within 11_Odd_World - the wrong pyre-village is owned by player-character at start. I haven't investigated where these coords are supplied.

Dugy:
If it simply sets the value that is subtracted from the upkeep, there is no point to give it an upper limit.

Also, I have fixed the village problem. I haven't noticed the added comment before.

PR #11, opened 2016-12-08
village support 1000 in scenario5

MPagel:
In scenario 5 cfgs, where `{DISABLE_UPKEEP 1}` appears, changed side 1 (Efraim) to have the following values
```
        gold=0
        income=0
        village_gold=0
        village_support=1000
```
except in Akula's place where `gold=200`

Note that if any scenarios don't have a village owned by Efraim at the start, the old `DISABLE_UPKEEP` code may still be needed. A "workaround" (instead of `DISABLE_UPKEEP`) on such maps would be to create a village "off map". Then that village is assigned to Efraim, shrouded, and never revealed. This would be needed to prevent teleportation to the site as a "safe heal" location.

### DIFF
--- a/scenarios5/08_Entrance.cfg
+++ b/scenarios5/08_Entrance.cfg
@@ -61,12 +61,13 @@
         side=1
         controller=human
         recruit=Skeleton,Skeleton Archer,Revenant,Deathblade,Bone Shooter,Ghost,Shadow,Wraith,Chocobone
-        gold=0
-        income=0
         shroud=yes
         team_name=good
         user_team_name=_"Good"
-        village_gold=1
+        gold=0
+        income=0
+        village_gold=0
+        village_support=1000
     [/side]
     [side]
         generate_name=yes
--- a/scenarios5/09_Northwestern_Storeroom.cfg
+++ b/scenarios5/09_Northwestern_Storeroom.cfg
@@ -61,12 +61,13 @@
         side=1
         controller=human
         recruit=Skeleton,Skeleton Archer,Revenant,Deathblade,Bone Shooter,Ghost,Shadow,Wraith,Chocobone
-        gold=0
-        income=0
         shroud=yes
         team_name=good
         user_team_name=_"Good"
-        village_gold=1
+        gold=0
+        income=0
+        village_gold=0
+        village_support=1000        
     [/side]
     [side]
         no_leader=yes
--- a/scenarios5/10_Guards_Quarters.cfg
+++ b/scenarios5/10_Guards_Quarters.cfg
@@ -68,11 +68,12 @@
         controller=human
         recruit=Skeleton,Skeleton Archer,Revenant,Deathblade,Bone Shooter,Ghost,Shadow,Wraith,Chocobone
         gold=0
-        income=0
         shroud=yes
         team_name=good
         user_team_name=_"Good"
-        village_gold=1
+        income=0
+        village_gold=0
+        village_support=1000
     [/side]
     [side]
         no_leader=yes
--- a/scenarios5/11_Northern_Guardian_Room.cfg
+++ b/scenarios5/11_Northern_Guardian_Room.cfg
@@ -57,12 +57,14 @@
         side=1
         controller=human
         recruit=Skeleton,Skeleton Archer,Revenant,Deathblade,Bone Shooter,Ghost,Shadow,Wraith,Chocobone
-        gold=0
         income=0
         shroud=yes
         team_name=good
         user_team_name=_"Good"
-        village_gold=1
+        gold=0
+        income=0
+        village_gold=0
+        village_support=1000        
     [/side]
     [side]
         generate_name=yes
--- a/scenarios5/12_Amplificator1.cfg
+++ b/scenarios5/12_Amplificator1.cfg
@@ -71,12 +71,13 @@
         side=1
         controller=human
         recruit=Skeleton,Skeleton Archer,Revenant,Deathblade,Bone Shooter,Ghost,Shadow,Wraith,Chocobone
-        gold=0
-        income=0
         shroud=yes
         team_name=good
         user_team_name=_"Good"
-        village_gold=1
+        gold=0
+        income=0
+        village_gold=0
+        village_support=1000
     [/side]
     [side]
         no_leader=yes
--- a/scenarios5/13_Eastern_Storeroom.cfg
+++ b/scenarios5/13_Eastern_Storeroom.cfg
@@ -71,13 +71,14 @@
         side=1
         controller=human
         recruit=Skeleton,Skeleton Archer,Revenant,Deathblade,Bone Shooter,Ghost,Shadow,Wraith,Chocobone
-        gold=0
-        income=0
         shroud=yes
         team_name=good
         user_team_name=_"Good"
-        village_gold=1
-    [/side]
+        gold=0
+        income=0
+        village_gold=0
+        village_support=1000
+[/side]
     [side]
         no_leader=yes
         side=2
--- a/scenarios5/14_Armoury.cfg
+++ b/scenarios5/14_Armoury.cfg
@@ -69,12 +69,13 @@
         side=1
         controller=human
         recruit=Skeleton,Skeleton Archer,Revenant,Deathblade,Bone Shooter,Ghost,Shadow,Wraith,Chocobone
-        gold=0
-        income=0
         shroud=yes
         team_name=good
         user_team_name=_"Good"
-        village_gold=1
+        gold=0
+        income=0
+        village_gold=0
+        village_support=1000
     [/side]
     [side]
         generate_name=yes
--- a/scenarios5/15_Amplificator2.cfg
+++ b/scenarios5/15_Amplificator2.cfg
@@ -70,12 +70,13 @@
         side=1
         controller=human
         recruit=Skeleton,Skeleton Archer,Revenant,Deathblade,Bone Shooter,Ghost,Shadow,Wraith,Chocobone
-        gold=0
-        income=0
         shroud=yes
         team_name=good
         user_team_name=_"Good"
-        village_gold=1
+        gold=0
+        income=0
+        village_gold=0
+        village_support=1000
     [/side]
     [side]
         no_leader=yes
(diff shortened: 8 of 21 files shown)


Dugy:
You need to set income to -2 if you want the side to receive nothing. If you set it to zero, the side still receives 2 gold per turn because that is some sort of basic income. Have you actually verified if your code is working?

I don't understand your concerts about owning a village on the start. Why do you need it?

MPagel:
true that.  another option would be to set `base_income` in `[game_config]`
https://wiki.wesnoth.org/GameConfigWML

but that may require a change to all non-affected scenarios.  I see code references to `{NO_INCOME}` but can't find a `#define` for that in your code, the wesnoth/wesnoth repo or the WML docs

MPagel:
re: requirement for a starting village.  As the support is 1000 per village, if you have no villages you have no support.  If you don't capture a village that first turn, you'll backslide credits equivalent to the character-levels of non-loyal/hero units. I didn't see a way to tie support to a character (e.g. Efraim) rather than a village owned by the side.

Dugy:
{NO_INCOME} just sets the income to -2 to offset base_income. I can't remember where is it defined, but I am certain that I haven't written it. It doesn't matter if it's set as game_config or in the side's definition, but I think that the latter is more transparent.

I see why you need the village now. I do not recommend doing that, it is quite a bug prone workaround I would prefer not to resort to it. Using the village_income without removing the DISABLE_UPKEEP macro hides the weird changes you have noticed in most cases and isn't error prone.

Also, please commit all the changes you want to a single PR, I don't want to have the history spammed with countless PRs with almost identical changes. One change should have one commit or at least one branch.

MPagel:
Yes, I had planned on committing more at a time, but I figured one scenario set at a time was a reliable way to handle it. But I shall close this pull request (after your next comment), and work in my copy of the repo and submit all-at-once. Perhaps I'll even work with awk or sed to make these changes en masse.  Would you prefer that I use `income=-2` or `{NO_INCOME}`? I think `{NO_INCOME}` is more intuitive and appears to magically work, wherever it is defined. `income=-2` would however be marginally quicker, as the macro doesn't need to be found and expanded. I "vote" for clarity of intention (and human readability of code) with `{NO_INCOME}`.

Dugy:
You don't need to close the pull request, you can just add some more commits into it. But it's up to you.

I prefer income=-2 to {NO_INCOME}, because the latter doesn't clearly state what does it to with village income and it's faster when loading (the number of uses isn't trivial and LotI's loading time is already quite suboptimal).

MPagel:
sounds good. I'll leave open and comment when I'm ready for your review/merge. Do you want to remove/comment out the disable upkeep code for scenarios where you start out with (a) village(s), or keep it in there as a failsafe? Or conditionally execute if villages (owned)<1, as I don't believe the macro will expand if the conditional block is not met?

Dugy:
The conditional thing you're proposing would not work.

I don't think there is a need for any comments about it in code (except one somewhere).

MPagel:
You're right that the conditional thing wouldn't save any time in scenario load, etc.  

From https://wiki.wesnoth.org/Wml_optimisation

> When Wesnoth loads scenario, all macroses are substituted with their meaning.

It might be possible to replace or augment the `{DISABLE_UPKEEP}` macro call with a [fire_event] where the event itself has the following if condition block to process the rest of `DISABLE_UPKEEP` code. the if condition would look like
```
[if]
   [variable]
      name=side_number
      equals={SIDE}
   [/variable]
# count villages for side desired
   [then]
      [store_villages]
         variable=owned_villages
         [filter]
            [if]
               [variable]
                  name=owner_side
                  equals={SIDE}
               [/variable]
            [/if]
         [/filter]
      [/store_villages]
# Now compare
      [if]
         [variable]
            name=owned_villages.length
            equals=0
         [/variable]
         [then]
# rest of old {DISABLE_UPKEEP} macro
# alternatively a FIRE_EVENT to do equivalent save/recall
            [event]
               name=side turn 
               first_time_only=no
               [store_gold]
                  side={SIDE}
                  variable=upkeep_mitigation_gold_container_{SIDE}
               [/store_gold]
            [/event]
            [event]
               name=turn refresh
               first_time_only=no
               [modify_side]
                  side={SIDE}
                  gold=$upkeep_mitigation_gold_container_{SIDE}
               [/modify_side]
               [clear_variable]
                  name=upkeep_mitigation_gold_container_{SIDE}
               [/clear_variable]
            [/event]
         [/then]
      [/if]
  [/then]
[/if]
```
This is, of course, a longer macro code. However, it should not look up gold if the side has a village, so might be quicker to process (then again, filtering villages and counting the length of the array might take longer). And it's possible there could be, as I said, a fire event coded that does the guts only if the village count condition is met. I had previously assumed that there was a villages_owned array automatically created for each user.

re: "comment out" - this is an English (American?) colloquialism that means to place a comment mark before the code to prevent its execution.  So the line would look like `#   {DISABLE_UPKEEP 1}`. I agree that the first time we do the income -2 bit, there should be a brief comment.  

P.S. I found `{NO_INCOME}`: https://www.wesnoth.org/macro-reference.xhtml#NO_INCOME

Dugy:
First, I'd like to apologise for skimming through your reply without reading it properly.

Longer definition is not a problem if it's possible to use the code in every scenario (so that it would be protected from activating in scenarios where it's not wanted). All stuff in global_events.cfg is loaded in memory only once and used in every scenario. You could define the macro as a fire_event that calls an event that is always present. But the execution time in every turn probably does matter.

I suggest to use both approaches at once, village_gold to to have it work well and DISABLE_UPKEEP as a failsafe for the case if player has no villages. Just add the village_gold thing and keep DISABLE_UPKEEP there.

And I agree that some sort of script could do that far easier, should I write one or you will do it?

MPagel:
alternatively/additionally, {DISABLE_UPKEEP} should only need to be called in scenarios where {STARTING_VILLAGES {SIDE}} isn't present (with the possible exception of scenarios with lava destruction or scenarios where you made that call but didn't put any villages within the radius)

https://www.wesnoth.org/macro-reference.xhtml#STARTING_VILLAGES

Dugy:
I may happen that enemies occupy all your villages. It may happen easily in scenarios where you have to get across the map (and you are passing through areas with no villages or are close to a defeat). DISABLE_UPKEEP should replace incorrect behaviour with small visual oddities. Like I said, a failsafe.

MPagel:
> And I agree that some sort of script could do that far easier, should I write one or you will do it?

[filesToMod.txt](https://github.com/Dugy/Legend_of_the_Invincibles/files/641695/filesToMod.txt) is the output of list of `grep -r -i -l "{DISABLE_UPKEEP 1}" .`.  I'm heading off to bed now, so feel free to take over.

MPagel:
its frustrating that values in the top status bar aren't directly accessible to WML

https://wiki.wesnoth.org/LuaWML/Tiles#wesnoth.get_villages

has an example
```
-- How many villages do we have on our map?
v = #wesnoth.get_villages()
```
but the "we" means the total count on the map? not the number owned by "we" the player?  If former, it should read "how many villages are on the map", if latter should read something like "how many villages do we own on the map." Ambiguities...urgh.

MPagel:
in wesnoth/src/team there are the following code snippets
header
```
bool owns_village(const map_location& loc) const
{ return villages_.count(loc) > 0; }

/** Calculate total support capacity, based on support_per_village. */
int support() const { return villages_.size()*village_support(); }
```
cpp
```
# MPagel github note set: owns_a_village = FALSE [or add add'l parameter to call stop_after_one]
for (const config &v : cfg.child_range("village"))
	{
		map_location loc(v);
		if (map.is_village(loc)) {
			villages_.insert(loc);
# MPagel github note: owns_a_village = TRUE
# and/or break from for loop if stop_after_one==TRUE
		} else {
			WRN_NG << "[side] " << current_player() << " [village] points to a non-village location " << loc << std::endl;
		}
}
```

Dugy:
wesnoth.get_villages() returns a list of all villages, regardless of owner.

Are you trying to create a lua-based function that would do the village-owner-or-disable-upkeep thing? The only advantage would be a slightly faster execution.

MPagel:
well...screwballs.  look what I found in  wesnoth/src/settings.cpp
```
int get_village_support(const std::string& value)
{
	return lexical_cast_in_range<int>(value, 1, 0, 4);
}
```
so, we would need to figure out a way to bypass that. sorry for the gigantic waste of time.

Dugy:
I don't get what you mean. That function looks like something that reads a string and returns a number. It may just be passed the value set and output the number. Have you tried if the village_support thing really doesn't work if the player has some units?

MPagel:
nope...still untested, but it looks to me that that's the underpinning code that vets your config files for valid values. I read it as default is 1, the value should be between 0 and 4.

Dugy:
I have no idea what that function does, but I thought it reads from the 1st until the 5th digit (indexes 0 to 4, C++ indexes from 0). Your interpretation makes more sense. Why don't you try it out?

MPagel:
@Vultraz suggests
```
[event]
    name=prerecall
    first_time_only=no

    [modify_unit]
       [filter]
            id=$unit.id
       [/filter]

       upkeep="full"
    [/modify_unit]
[/event] 
```
and follows that up with 
>Upkeep is not the same as having the Loyal trait. A unit with the Loyal trait also has 0 upkeep, but one can set upkeep to either loyal or free without having the Loyal trait.

Vultraz:
For the record, a filter will need to be added to that event to only affect the desired units, or else it will affect any unit being recalled by any side. I didn't write it with one since I don't know what said desired units are.

Dugy:
That's a good idea guys.  Having all units get that silent loyal property in a unit_placed event (on 1.13.3 and later only) could do it. If that macro's not there, upkeep could be set back to full after a check if the unit's not supposed to be loyal.

Because there are too many other ways the units can enter the game, some of them have no event attached (such as plague), so it would not work on versions before 1.13.3. A little problem is that I haven't made LotI work properly on 1.13 yet.

MPagel:
Re: Plague.  Standard plague results in a level 0 unit, which is functionally loyal until advancement. Custom plagues (e.g. straight to Reverent) are already firing WML events, so could have a function/macro call inserted.  The same, I presume, with finding units in villages or random allied characters "deciding" to join your team.

Dugy:
No plague units fire an event upon them (not even custom plague). Fixing up all scenario code when units are added is not a good way.

Dugy:
Yes, it has been solved.

PR #12, opened 2016-12-19
Fix broken AI in 1.13.x

Battlecruiser:
This fixes NPEs in Wesnoth 1.13.x


Dugy:
The LotI AI is a part of 1.13 code, named as fast micro ai. All that needs to be done is to deactivate the AI definition altogether and let it use the default micro ai.

I suspect that your fixes would break compatibility with 1.12. I can't accept it if it does.

matsjoyce:
Since we are not targeting 1.14 which has the new micro AIs by default, we can drop the ai code and this PR?

Dugy:
Yes.

PR #22, opened 2017-04-08
Fix Abaddon missing his message

matsjoyce:
Abaddon currently does not have his message appear because his unit name is `Abaddon-<random number>`. I have tested this and it fixes the problem.


Dugy:
Thanks.

PR #27, opened 2017-05-15
Track down missing amlas and finish renaming started in cadbf10

matsjoyce:
I adjusted my script to print out missing amlas that it couldn't find and this is the result. I'm not 100% sure if I changed the correct thing in each case, but this fixes all of the cases the seem unintended AFAIK. Also finished renaming the Theophrastus Redivivus.


matsjoyce:
Well, I tested it with my script as testing it with wesnoth could take a while (or lots of debug). Does that count? :grin:

Dugy:
Yes. Merging.

PR #28, opened 2017-05-16
Fix ogg.ogg extensions

matsjoyce:
I noticed some scenarios had a possible broken SCENARIO_MUSIC tag. This should fix all the cases.


Dugy:
Merged. Thank you.

PR #32, opened 2017-05-31
minor Efraim's iceball description fix

arodiss:


Dugy:
Do you want commit access? Making a pull request for changing one word must be annoying.

Dugy:
It will break translations. Changes of text can almost never break anything else.

konecnyjakub:
They can break the whole campaign if you are not careful: konecnyjakub/Saga_of_James@86871545. I was having really hard time finding the mistake (because Wesnoth was quite misleading me and I do not have syntax highlighting for WML in my editor).

Dugy:
I believe that one would be noticed very fast.

konecnyjakub:
On GitHub, sure. It does some highlighting for WML in commits and you knew which commit introduced that bug. But if you forget to test every little change, you can be stuck for months (I was).

Dugy:
A forgotten quotation mark can be found quickly even without a highlighter. Simply remove all files from the folder, then put them back one by one till you find the file, then remove all its contents, then put a half of it back, see if the error was there, then put the half of the half back and so on, until you pin down the exact line.

PR #35, opened 2017-05-31
[Requires review] Simplify binary search and try to plug all the holes

matsjoyce:
Improves on the sorting of items that was added before. I've gone though and tried to update everywhere that `item_storage` is added to, as currently the item list slowly goes out of order due to unpatched modifications. I have tested all the changed cases except `scenarios10/14_End_of_the_World.cfg` as I don't have a replay of that ATM. It's similar enough to `scenarios8/14_Finally_Together.cfg` that the code should be correct, though.


Dugy:
All right. Thanks.

PR #37, opened 2017-05-31
More inversion corrections

matsjoyce:
I adjusted the wiki script to detect errors like that in #36 and this was what it found.


Dugy:
Good job.

PR #38, opened 2017-06-01
Collapse items of the same type when viewing item storage

matsjoyce:
Changes:

To:

This shouldn't change translatability of anything.


Dugy:
Thanks.

PR #40, opened 2017-06-26
Fixed error in walkthrough

arodiss:
For some reason GitHub does not highlight it, but I replaced text
... there will be a trader in a town near the **northern** border ...
with
... there will be a trader in a town near the **southern** border ...

This is scenario 8.8


Dugy:
Thank you.

PR #51, opened 2017-11-06
Fix backstab overriding

matsjoyce:
This patch fixes backstab overriding. However, is causes the following problem:

The Exterminator has [an advancement](https://wiki.wesnoth.org/LotI_Unit_Advancements#Able_to_Charge_at_Enemies_when_Backstabbing_if_Selected_.E2.80.93_murderous_frenzy) that is a clone of the normal attack with charging backstab added. Since the normal attack has backstab, the new attack is basically a weaker version of the original attack with no new specials, and so is useless. Either:

 - Charging backstab needs a bigger increase in damage
 - Or it should not be overridden, so it stacks with other backstab specials

### DIFF
--- a/utils/stats.cfg
+++ b/utils/stats.cfg
@@ -3987,21 +3987,31 @@
                         variable=advanced.attack[$i].specials.damage[$j].id
                         [case]
                             value=charging backstab
-                            {VARIABLE backstab 1}
+                            {VARIABLE new_backstab 1}
                         [/case]
                         [case]
                             value=backstab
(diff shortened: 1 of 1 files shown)


matsjoyce:
@Dugy What are we going to do with the Exterminator advancement?

Dugy:
I am not quite sure why have I given him such a shitty advancement. Maybe because it did stack with the usual backstab. I think that it should give a 100% extra damage or something.

PR #55, opened 2017-11-07
Remove female variation from Efraim

matsjoyce:
@Dugy I can't find any uses of this, so can it be removed? Reported here: https://forums.wesnoth.org/posting.php?mode=reply&f=8&t=32384#pr619466


Dugy:
Delly used to be identical to Efraim, just female. I eventually changed her into a custom unit type, leaving the old female version there to keep old save files working. It was years ago, there's no need for that any more.

PR #82, opened 2018-04-18
Fix grammar

soliton-:
Not so sure about the preach thing. Still sounds weird to me.


Dugy:
What preach thing?

soliton-:
The second change in the commit.

Dugy:
I have looked at example usage in a dictionary and it seems that _preach to us_ is correct.

PR #97, opened 2018-10-19
Fix typo in 15_Long_Way_Home.cfg

edwardspec:
"they had had" -> "they had"


Dugy:
I think it was a grammar error rather than a typo, but it doesn't matter. Merging. Thanks for contributing.

PR #131, opened 2018-12-06
Gladiators: prevent Woodland Cloak (item 34) from being dropped

edwardspec:
This item is useless in Gladiators, because the arena has no forests.
(all bonuses provided by the Woodland Cloak are forest-specific)


Dugy:
Sorry, the defines don't work that way. If you define something in a scenario, it won't make it define if and only if the scenario is run.

Or did something in Wesnoth change and enable it?

edwardspec:
Indeed, define would have affected other multiplayer scenarios.
Replaced with a variable multiplayer_gladiators=1, which is set at the prestart event of Gladiators.

Dugy:
This still isn't a very useful approach. If all campaigns that use LotI as resource had this kind of exception for every item related to something that doesn't happen to appear there, the file would be a huge tangled mess from all those exceptions.

I suggest these alternative solutions:
1. Add support for a variable that holds id numbers of items that cannot spawn
2. Create a similar variable that would exclude all terrain-specific and enemy-specific items (I am sure Woodland Cloak isn't the only useless item in that scenario)
3. Set those drop lists as variables that can be replaced in specific scenarios
4. Let it be

edwardspec:
I'll investigate solution 1.

Solution 3 is simpler, but has a drawback: newly added items will never drop in existing (third-party) scenarios that already use this feature (until they are updated).

Solution 2 will likely require the same code to be implemented as solution 1.
Plus some third-party scenarios may want to exclude more than can be covered by these criteria.

Dugy:
I think that the easiest way to implement it is to try to pick an item three times and retry if it is on the list.

If you need some question answered, ask.

edwardspec:
Quick testcase:
```
diff --git a/utils/item_list.cfg b/utils/item_list.cfg
index 545590e..2a21c33 100644
--- a/utils/item_list.cfg
+++ b/utils/item_list.cfg
@@ -42,7 +42,7 @@
             spear=91,91,92,92,93,94,95,140,140,140,150,150,175,180,198,198,231,303
             amulet=87,87,87,17,17,17,12,12,2,123,134,134,134,161,161,164,174,178,183,234,244,244,264,313,322
             ring=35,35,35,33,33,21,21,21,73,73,96,96,83,18,9,116,116,122,19,19,154,154,120,162,162,166,166,177,186,194,194,194,266,266,269,283,284,321,335,335,335
-            cloak=27,34,56,79,85,121,131,133,176,192,215,232,252,258,265,270,273,281,289,294,294,300,327
+            cloak=27,34,56
             book=39,39,89,89,26,26,20,20,13,13,8,8,5,5,101,101,102,103,103,104,104,105,124,181,216,296,297,298,304,336
             potion=43,43,43,43,43,43,43,43,43,43,42,42,32,32,16,16,38,38,1,106,106,106,106,130,130,144,144,155,155,193,193,201,204,204,205,205,205,205,205,205,243,243,246,259,275,295,315,330
             gold=24,37,37,44,44,44
@@ -50,7 +50,7 @@
         [/value]
     [/set_variables]
 
-    {VARIABLE_OP item_kind rand (armour,weapon,other)}
+    {VARIABLE_OP item_kind rand (other)}
     [switch]
         variable=item_kind
         [case]
@@ -63,7 +63,7 @@
         [/case]
         [case]
             value=other
-            {VARIABLE_OP item_sort rand (amulet,amulet,ring,ring,cloak,cloak,book,potion,potion,potion,potion,potion,gold,gold,gold,gold)}
+            {VARIABLE_OP item_sort rand (cloak)}
         [/case]
     [/switch]
     [lua]

```

... and then try the following in Gladiators.cfg:

Test 1. No droppable_items_blacklist. 
(can drop Nocturnal Cloak, Woodland Cloak or Cloak of Retribution. Shouldn't display any errors or warnings)

Test 2. `{VARIABLE droppable_items_blacklist 34}`
(can drop Nocturnal Cloak or Cloak of Retribution)

Test 3. `{VARIABLE droppable_items_blacklist 27,34}`
(should only drop Cloak of Retribution)

Test 4. `{VARIABLE droppable_items_blacklist 27,34,56}`
(can drop any of those three, because blacklist that excludes ALL items is ignored by a sanity check)

edwardspec:
>     1. It doesn't handle the possibility when someone blacklists all items of a certain type.

It does. Please note the following line of the code:
```lua
options = options:gsub("^,", ""):gsub(",$", ""):gsub("^$", droppable_items)
```

If all items were blacklisted, then `options` will be an empty string, which will match `^$`.
If this happens, it gets replaced with the value of `droppable_items` - all items of this type (with nothing excluded).

edwardspec:
> 2. It expects to have some unobvious variables set around it, it should better be implemented as a new WML tag that has an argument dictating the item type that should be generated and the variable where the result should be saved

Will investigate this.

Dugy:
Very well. Thanks for the contribution.

PR #136, opened 2018-12-11
Implement "can unit A equip item B?" check in Lua

edwardspec:
This change has two purposes:
1) (short-term) loti_util_list_equippable_sorts() is much shorter
and easier to read than the WML conditionals.

2) (strategic)
Now that we know a list of "what kinds of items can unit A equip?",
we can potentially improve the "items on this unit" display by adding
placeholders for items that can be equipped - but are not yet equipped.

It's a common annoyance when you discover that your important unit
didn't have a cloak equipped, simply because you forgot to equip it,
and there was no way to quickly check.

In essence, I want to see the following:
sword: [none]
... in the inventory of units who can equip swords.

Furthermore, if/when such feature is implemented, players will know
that, for example, Paladin can equip a staff and that "torch" attack
can use "otherworldly essence" item, which is not self-evident.

Previously such feature was impossible, because all those WML checks
are local, so "can this unit equip a sword?" became known only when
a unit tried to pick one.
But loti_util_list_equippable_sorts() returns an array of sorts
and can be called from anywhere.

Note: utility function loti_util_list_attacks() is same as has_attack()
in titles.lua, and may replace has_attack() later.


edwardspec:
1100+ deleted lines :innocent:

Dugy:
Oh, very nice. Thanks for the contribution.

matsjoyce:
Niiice, this sort of change will really make stuff easier to bugfix. :+1:  If you are continuing this you might want to make a `lua` folder and put code into different files instead of just the monolithic `lua.lua`.

Dugy:
@matsjoyce Yes, it might soon become necessary.

PR #137, opened 2018-12-11
Replace deprecated macros like ABILITY_LEADERSHIP_LEVEL_5

edwardspec:
In Wesnoth, macros like ABILITY_LEADERSHIP_LEVEL_3 are now deprecated
and defined to be aliases of the standard ABILITY_LEADERSHIP.

This made Efraim (demigod) a ridiculously overpowered lvl 10 leader
(for adjacent lvl 4 units, the bonus to damage was +150%),
where only lvl 5 leadership was intended.

In [abilities.cfg], we now define a macro "ABILITY_LEADERSHIP_LEVEL x",
where x is the level.

We also remove definition of ABILITY_LEADERSHIP_LEVEL_6 (as duplicate)
and show "leadership (as level N unit)" as displayed name of ability.


Dugy:
Well done, I have not noticed this problem.

However, you have one mistake there. Translation doesn't work if it has a macro in it. 

Please replace:
`name= _ "leadership (as level {LEVEL} unit)"`
with:
`name= _ "leadership (as level " + {LEVEL} + _" unit)"`

Dugy:
1.14. There is no need to support earlier versions.

edwardspec:
Here we go.
(no more "deprecated macro" warnings in console when starting Wesnoth)

Issue #95, opened 2018-09-12
Replace redeem upgrade screen with a lua-based GUI


Dugy:
It would look very cool and could be easily backported to 1.10 to deal with the iPhone showstopper.

edwardspec:
Will investigate this.

Dugy:
Leave this to me, this is really annoying to learn because of bugs and strange behaviour and I have some experience in it.

edwardspec:
Too late... already wrote most of it.

Dugy:
Very good.

Though I would prefer removing that _provides_ section and listing the unavailable ones below the available ones with red title and a list of redeem advancements the user needs to get first.

And maybe the description could be on a second line.

Dugy:
I think that the problem with clickable red items could be easily fixed by adding a callback to selecting them that would disable to _accept_ button.

edwardspec:
I'm using wesnoth.show_menu(), it doesn't seem to support any callbacks.
https://wiki.wesnoth.org/LuaWML/Display#wesnoth.show_menu

It just takes an array of menu items and returns the index of selected one.

Dugy:
Are you sure that function was available on 1.10?

I thought you were using wesnoth.show_dialog(), which is what I always used for that, even on 1.10. It would also showing additional info, like number of advancements already taken and the amount of redeeems that will be needed to reach the next level.

edwardspec:
Yes, show_dialog() is present in Wesnoth 1.10: https://github.com/wesnoth/wesnoth/blob/1.10/src/scripting/lua.cpp

edwardspec:
Will replace wesnoth.show_menu() with wesnoth.show_dialog() (show_menu was added in Wesnoth 1.13.6).
Not much difference...

Dugy:
Good.

Now when you have wesnoth.show_dialog(), you can easily add callbacks. It's described at https://wiki.wesnoth.org/LuaWML/Display#wesnoth.show_dialog

I can finish it for you if you want, I was busy all day but now I am not.

edwardspec:
I will complete it myself (if you don't mind).
It's a good task for learning.

Dugy:
Good luck.

edwardspec:
Here is how it looks with `wesnoth.show_dialog()`:

![redeem-menu2](https://user-images.githubusercontent.com/1163218/49908150-4288be80-fe8a-11e8-9cb2-1aca75192c28.jpg)

[...]

PR #140, opened 2018-12-13
Replace redeem upgrade screen with a lua-based GUI 

edwardspec:
See issue #95.
Works for me (with both newly added unit in Gladiators and on real Demigod in the campaign - selected upgrade is indeed unlocked, and existing upgrades correctly affect the menu), but any additional testing would be very welcome.


Dugy:
I checked it and it seems to be all right. Thanks for the contribution. I will shoehorn it into the 1.10 version so than iPhone users would get their greatest pain fixed.

Issue #90, opened 2018-08-09
Images not showing in "market" menu


arobinson:
Current git "trunk" in use with Wesnoth 1.14.4 on MacOSX via steam. 

Issue is that items do not have images but rather &items/book3.png= for example

See screen shot

arobinson:
![screen shot 2018-08-09 at 10 22 42 am](https://user-images.githubusercontent.com/93213/43912092-8004a802-9bbe-11e8-9454-f798c20a9947.png)

arobinson:
Found in the Lotl1-Long Way Home scenario

Dugy:
I was able to replicate it on BfW 1.14.3 on Ubuntu via backports. I am not sure why it doesn't appear, it might be some change in 1.14 I don't know of. I will have to investigate more.

Dugy:
Fixed in #141

PR #141, opened 2018-12-13
Fix images not being shown in trading menus and "Choose your hero" menu in Gladiators

edwardspec:
This solves issue #90 (images not showing in "market" menu).


Dugy:
Thank you.

PR #146, opened 2018-12-14
Use icons of anti-ice/fire/arcane armor for "Resist [...]" upgrades

edwardspec:
Previously all "Resist [...]" upgrades used the same icon (generic
"leather armor" icon), which didn't look good in redeem menus.

There aren't any distinct anti-pierce/impact/blade armor images,
so their "Resist [...]" upgrades weren't changed (yet).

### DIFF
--- a/lua/redeem.lua
+++ b/lua/redeem.lua
@@ -87,17 +87,17 @@ known_ability_trees.redeem = {
 		requires = { "magic" }
 	},
 	resist_fire = {
-		image = "items/armour-fire.png",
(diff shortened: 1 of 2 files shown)


edwardspec:
Note: these images don't have borders (unlike other upgrade icons), but they look ok in upgrade/redeem menu nevertheless.

Dugy:
You can give the icon a standardised background by setting the image path from `items/armour-fire.png` to `attacks/blank-attack.png~BLIT(items/armour-fire.png)` and similarly for other icons.

Dugy:
All right. Thanks for the contribution.

PR #147, opened 2018-12-15
Reimplement DISABLE_UPKEEP: temporary unit.upkeep=loyal on placed units

### LINKED ISSUE #8: a potential fix for changing money in scenarios with DISABLE_UPKEEP
This "fix" may have unknown consequences, but might be worth testing. Note, my other comments in code were assuming LUA code, not WML.  Frustratingly they have different comment conventions.

```
#define DISABLE_UPKEEP SIDE
    [event]
        name=side turn end 
# will shift the storage of upkeep_mitigation_gold_container_{SIDE} to when the player pushes "end turn" which might be more reliable? Does the variable then need to be declared global for successive turns to have the value set? When in the sequence is the beginning-of-turn autosave created? is it after turn refresh? If before turn-refresh and the upkeep_mitigation_gold_container_{SIDE} is not saved to the autosave, this might result in the gold discrepancies seen.
        first_time_only=no
        [if]
            [variable]
                name=side_number
                equals={SIDE}
            [/variable]
            [then]
                [store_gold]
                    side={SIDE}
                    variable=upkeep_mitigation_gold_container_{SIDE}
                [/store_gold]
            [/then]
        [/if]
    [/event]
    [event]
        name=turn refresh
        first_time_only=no
        [if]
            [variable]
                name=side_number
                equals={SIDE}
            [/variable]
            [then]
                [modify_side]
                    side={SIDE}
                    gold=$upkeep_mitigation_gold_container_{SIDE}
                [/modify_side]
                [clear_variable]
                    name=upkeep_mitigation_gold_container_{SIDE}
                [/clear_variable]
            [/then]
        [/if]
    [/event]
#enddef
```

edwardspec:
Here is the approach:
1) units with loyal trait are unaffected,
2) every non-loyal unit that appears on the battlefield (regardless of
how it appeared there - [unit], recruit, recall, debug, whatever else)
is marked as upkeep=loyal.
Note: this is invisible (not shown on the unit, unlike "loyal" trait).
3) just in case some unit had some tricky upkeep value (e.g. upkeep=7),
we save the original unit.upkeep into unit.variables.original_upkeep.
4) in the "victory" event, (2) is undone, original upkeeps are restored
and unit.variables.original_upkeep is cleared.

This resolves #8 and #11.

### DIFF
--- a/utils/utils.cfg
+++ b/utils/utils.cfg
@@ -1046,41 +1046,55 @@ return leader_ai
         {CLEAR_VARIABLE side_{SIDE}_original_user_team_name,side_{SIDE}_original_team_name,NEUTRAL_SIDE_i,NEUTRAL_SIDE_side_store}
     [/event]
 #enddef
+
 #define DISABLE_UPKEEP SIDE
     [event]
-        name=side turn
+        name=unit placed
         first_time_only=no
-        [if]
+        [filter]
+            side={SIDE}
+        [/filter]
+        [filter_condition]
             [variable]
-                name=side_number
-                equals={SIDE}
+                name=unit.upkeep
+                not_equals=loyal
             [/variable]
-            [then]
-                [store_gold]
-                    side={SIDE}
-                    variable=upkeep_mitigation_gold_container_{SIDE}
-                [/store_gold]
-            [/then]
-        [/if]
+        [/filter_condition]
+        [modify_unit]
+            [filter]
+                find_in=unit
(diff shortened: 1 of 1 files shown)


edwardspec:
I also added standard village_gold=0 and income=-2 in [modify_side], because it's needed in literally every single scenario where we have DISABLE_UPKEEP.

Dugy:
Good idea.

PR #150, opened 2018-12-15
Add missing lua/debug.lua

edwardspec:
(wasn't added in efbe43fd1f847f40e1109a02bfc12665c65b2b43)


Dugy:
Oops, dammit. Thanks for committing it.

PR #156, opened 2018-12-18
Fix various minor typos, syntax errors, etc. spotted by looking at the wmllint log

edwardspec:
I analyzed the wmllint log and fixed everything that wasn't a false positive.
See commit descriptions for details on each fixed issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ before_script:
 
   # Apply pull request #3785 - required bugfix for wmllint.
   # Remove this line after https://github.com/wesnoth/wesnoth/pull/3785 gets merged.
-  - ( cd wesnoth && git fetch origin pull/3785/head:bugfix && git merge bugfix )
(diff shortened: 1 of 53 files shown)


edwardspec:
Travis log before: https://travis-ci.org/edwardspec/Legend_of_the_Invincibles/builds/469208433
Travis log after: https://travis-ci.org/edwardspec/Legend_of_the_Invincibles/builds/469310054

edwardspec:
The plan for later (not in scope of this pull request) is to teach wmllint not to complain about things that aren't a bug (e.g. we have `[/unit_type]` before `[unit_type]` in `GENERIC_AMLA` macro).

This is easy, because wmllint understands "ignore this issue" comments. For example, I will add such comment to `titles.cfg`, telling wmllint not to expect `variation_id` parameter in `[variation]` tag. There would only be a few of these comments, they will not affect code readability or anything.

Once wmllint has no false positives, I'll make Travis test to report failure if wmllint complained, and we could get those green/red circle icons to be automatically shown near commits that pass/fail a Travis test.

As with any automated testing, the idea is to get notified if some new change introduced a problem.

Issue #158, opened 2018-12-21
Redeem error: nil value


arobinson:
Screen shot: 
![screen shot 2018-12-20 at 8 44 33 pm](https://user-images.githubusercontent.com/93213/50323732-3dce9680-0498-11e9-977c-48f90ca0393b.png)

Save game before it happened:
[LotI2-Search for Elves Turn 6.gz](https://github.com/Dugy/Legend_of_the_Invincibles/files/2701424/LotI2-Search.for.Elves.Turn.6.gz)

Have Lethalia redeem the unit near her (I picked the entropy one). 

Error negates the gained advancement (in this case, cannot obtain the entropy skill)

arobinson:
Thanks

PR #159, opened 2018-12-21
Fix for issue 158

arobinson:


arobinson:
Function was being called with 1 parameter, but it takes 2

https://github.com/Dugy/Legend_of_the_Invincibles/issues/158

PR #163, opened 2018-12-25
Fix errors in 00_Tutorial caused by [store_unit]kill=yes before [update_stats]

edwardspec:
The bug in Tutorial was reported in the forum.

This fix has been tested by completing the Tutorial.

### DIFF
--- a/scenarios1/00_Tutorial.cfg
+++ b/scenarios1/00_Tutorial.cfg
@@ -292,19 +292,10 @@ Take this item?"
             [remove_item]
                 x,y=13,6
             [/remove_item]
-            [store_unit]
-                [filter]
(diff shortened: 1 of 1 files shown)


edwardspec:
Other scenarios (the ones that call update_stats) seem to be ok.

edwardspec:
Too bad Wesnoth scenarios can't be automatically tested (like browsers with Selenium, where you just write a script telling them where to click and then you can analyze what they show on the screen).

It would have been nice if a script could play all scenarios by itself (with some pre-programmed guidance on what to do), and then report all the errors, warnings and failed assertions. Would instantly catch things like this.

Dugy:
Thanks.

PR #164, opened 2018-12-26
Move crafting menu to [event]name=crafting_menu (to call from Lua)

edwardspec:
From the "inventory" branch.
I tested that this correctly works with the existing "Items" menu.

Contents of `[command]` were moved into `[event]` (except the line `{VARIABLE to_remove -1}`, which is only needed for the Items menu). No changes except indentation. (-20 spaces at the beginning of each line)


Dugy:
Should I merge it onto master or do you need it as a branch?

edwardspec:
Please merge it.

The idea is to merge things that don't depend on the new Inventory dialog in separate pull requests, for easier checking and all.

> If I merge it, I might create a lua-based GUI for crafting

Good.
Then we should probably merge [lua/items.lua] file too (the library for working with items - https://github.com/edwardspec/Legend_of_the_Invincibles/blob/inventory/lua/items.lua ), as it can be useful for the crafting code. Methods like `loti.item.on_the_ground.add()`, `loti.item.type[number]`, etc.

Dugy:
All right, merged.

Is lua/items.lua a separate file or does it contain stuff specific for the items menu that isn't finished yet? I can do the crafting later, without messing up the repo with weird order of commits, I have other projects I can work on now.

edwardspec:
> Is lua/items.lua a separate file

Yes. It was intentionally written to be independent from the items menu.
Will double-check all functions and submit a pull request.

>  without messing up the repo with weird order of commits

That's what "git rebase" is for.
I'll rebase the "inventory" branch before creating a pull request for it, so that it wouldn't have changes in lua/items.cfg (because this file is already present).

Dugy:
Okay.

PR #166, opened 2018-12-26
Add [lua/items.lua] library for manipulating LotI items in Lua

edwardspec:
All methods are documented (purpose, parameters and return values).
Complete list:

```lua
loti.item.storage.add = function(item_number, crafted_sort)
loti.item.storage.remove = function(item_number, crafted_sort)
loti.item.storage.list_items = function(item_sort)
loti.item.storage.list_sorts = function()
loti.item.on_unit.list = function(unit)
loti.item.on_unit.find = function(unit, item_sort)
loti.item.on_unit.add = function(unit, item_number, crafted_sort)
loti.item.on_unit.remove = function(unit, item_number, crafted_sort)
loti.item.on_the_ground.add = function(item_number, x, y, crafted_sort)
loti.item.on_the_ground.remove = function(item_number, x, y, crafted_sort)
loti.item.on_the_ground.list = function(x, y)
loti.item.util.undress_unit = function(unit)
loti.item.util.take_item_from_unit = function(unit, item_number, crafted_sort)
loti.item.util.get_item_from_storage = function(unit, item_number, crafted_sort)
```

There is also a pseudo-array `loti.item.type[item_number]` for fetching [object] WML table from the item_list by item_number.

Note: `unit` parameter is always a Lua unit object, not a WML table. These methods work directly on the unit, they don't make temporary copies of it.

Parameter `crafted_sort` is only needed if `item_number` refers to a crafted item. It can be omitted if we are working with non-crafted item.

### DIFF
--- a/lua/items.lua
+++ b/lua/items.lua
@@ -212,6 +212,14 @@ loti.item.on_unit.find = function(unit, item_sort)
 	return nil -- Not equipped
 end
 
+-- Internal: call update_stats on Lua unit object.
+local function update_stats(unit)
+	local updated = wesnoth.update_stats(unit.__cfg)
+	if unit.valid == "map" then
+		wesnoth.put_unit(updated)
+	end
+end
+
 -- Add one item to the unit.
 -- Optional parameter "crafted_sort" changes the item_sort of item (only for crafted items).
 loti.item.on_unit.add = function(unit, item_number, crafted_sort)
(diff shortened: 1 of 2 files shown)


Dugy:
The `loti.item.type` metamethod is implemented with excessive complexity. It should be an array created during one run through all item types.

The rule whether an `[object]` is an equipped item is that its `sort` property is set, there is no rule whether it should be `silent` or not.

The `loti.util.undress_unit()` function has excessive complexity, because it calls `update_stats()` after removing every single item. It should just run through the array and remove anything that should be remove.

I can fix it, but it would render the rest of your code incompatible.

edwardspec:
Will fix this shortly.

edwardspec:
> The rule whether an [object] is an equipped item is that its sort property is set, there is no rule whether it should be silent or not.

I removed the check for item.silent.
Still, by this definition the fire when being incinerated is an equipped item, so are the hp effects of Foul Potion.
(thy are both added as [object] and have `sort` key).

edwardspec:
> The loti.item.type [...] should be an array created during one run through all item types.

Fixed. Now `item_list` is only loaded once.
Note that items.lua is loaded way before `item_list` is populated, so `__index` is still needed to prepare `loti.item.type` on the first access.

edwardspec:
Will also implement `can_remove()` method or something,
so that the checks "is this a potion or a book?" wouldn't get duplicated in other code. (we now need to check this in `undress_unit()`, in Inventory dialog, etc.)

Dugy:
Some effects are marked with some sort of fake `sort` so that they would be automatically removed on level-up. Therefore anything that _contains_ the string `potion` is not unequippable. It's not very well-thought.

This approach could be replaced by a `remove_on_levelup=yes` key or something.

edwardspec:
Should be ok now, please have a look.

> This approach could be replaced by a remove_on_levelup=yes key or something.

Also Orbs could use "orb=yes".
This would break backward compatibility (for players who have these items in their saves - without remove_on_levelup key), so let's do this later.

Dugy:
There is no difference between orbs and books, so I don't think there is a reason to make some difference between them.

edwardspec:
I see. I thought orbs were the regular items and could be unequipped. (but I just checked and they can't)

PR #170, opened 2018-12-27
items.lua: exclude orbs from list_regular(), they can't be unequipped

edwardspec:
Due to https://github.com/Dugy/Legend_of_the_Invincibles/pull/166#issuecomment-450121239

### DIFF
--- a/lua/items.lua
+++ b/lua/items.lua
@@ -180,16 +180,9 @@ loti.item.on_unit.list_regular = function(unit)
 	local items = {}
 
 	for _, item in ipairs(loti.item.on_unit.list(unit)) do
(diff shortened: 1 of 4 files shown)


Dugy:
There is still a check for the `silent=yes` key, which should play no role.

edwardspec:
Removed silent check. Also removed those fake sorts, replaced them with id= key.

edwardspec:
Rebased for cleaner Git history after merging.

Dugy:
Be careful with the `id=` key, it is somehow used to prevent the same object from being applied more than once. In this case, it doesn't matter, but it could cause serious problems if it was used in the item system.

PR #185, opened 2018-12-28
Improvements to the new crafting dialog

edwardspec:
See the list of commits for details.


Dugy:
It's definitely better than what I have done, but if I change from crafting armours to weapons, the listbox to choose between armour and weapon shrinks (I don't know what can be done about it) and I can't find some recipes on the list. If I click on a blank recipe (below the last one), it crashes due to trying to work with a nil variable.

Do you want to fix it or will you leave it to me?

edwardspec:
Will investigate this shortly.

Issue #191, opened 2018-12-28
Crafting in Tutorial scenario should look like new crafting dialog


edwardspec:
In the Tutorial, Efraim has to craft 1 sword from 3 obsidians. Anything else is unnecessary.

I suggest making a fake dialog: copy get_dialog() from crafting.lua to get the same look,
but delete everything unnecessary:
1) show this one sword only,
2) not show Transmuting button,
3) don't calculate anything. Always show 3 obsidians and 0 everything else.
4) add same dialog options (e.g. Delenia saying that she doesn't remember any armour recipes if player clicks on "Armour", "Don't give up" when player tries to close the dialog, etc.).

This can be saved in `tutorial.lua` and then included in Tutorial only via
```lua
[lua] code = << wesnoth.require("tutorial") >> [/lua]
```

Dugy:
How are you dealing with this issue in your items GUI system?

edwardspec:
Differently. (such approach is not needed for Crafting window, because it's way more simple)

There would be a real Inventory dialog (not a fake copy of it).
However, it provides (or rather "will provide", the syntax is not yet determined) some Lua functions (that `tutorial.lua` can call) that allow a certain degree of modding.
Installing different callbacks for buttons, changing text, hiding unnecessary buttons, etc. That sort of things.

Dugy:
Okay, then. Just avoid duplicity wherever you can. The current tutorial is a really ugly heap of copy/paste code.

edwardspec:
> copy get_dialog() from crafting.lua to get the same look

Actually even this is not needed, only preshow() will be different.

Dugy:
Can't you just call those functions directly from `crafting.lua`?

edwardspec:
Yes, that's what I meant.
Tutorial can call get_dialog() from crafting.lua (because WML of the dialog is exactly the same) and write preshow() itself (because it's completely different).

Dugy:
Okay, then.

edwardspec:
Please clarify whether I should handle this task or leave it to you. (so that we wouldn't accidentally write it twice)

Dugy:
Handle it yourself. I am fixing those bugs that were found in my code and it's getting late here.

edwardspec:
Ok.

PR #194, opened 2018-12-29
Tutorial: use new crafting dialog during the crafting lesson

edwardspec:
See issue #191 ("Crafting in Tutorial scenario should look like new crafting dialog").


Dugy:
Do you want me to merge it?

edwardspec:
Please do, and try it out too.
(it's the second step of the tutorial)

Dugy:
It seems to be fully operational.

Maybe the transmuting button could be disabled rather than absent, it would make the layouts more similar.

Issue #190, opened 2018-12-28
New transmute dialog does not update the gem count


arobinson:
In the new transmute dialog, there is no current gem count, so there is no way to know how many gems are left (to answer the question, "should I keep transmuting?").

PR #197, opened 2018-12-30
Update the gem counts in Crafting dialog after each Transmute

edwardspec:
This fixes https://github.com/Dugy/Legend_of_the_Invincibles/issues/190


Dugy:
Huh, I didn't know it can be done so easily.

PR #199, opened 2018-12-30
Fix update_stats() not reducing defense of crafted non-armours

edwardspec:
(crafted gauntlets were adding the same defense as crafted armours)

### DIFF
--- a/lua/items.lua
+++ b/lua/items.lua
@@ -256,11 +256,6 @@ loti.item.on_unit.add = function(unit, item_number, crafted_sort)
 		end
 
 		item.sort = crafted_sort
-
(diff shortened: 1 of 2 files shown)


edwardspec:
Question:
should this only affect the bonus "Increases physical resistances by N%",
or should all other resistance bonuses be reduced to 1/3 too? (resist_blade, resist_impact, etc.)

Dugy:
Only defence, not the other resistance bonuses. I was adding the crafting recipes with that in mind.

However, if you think it would be better if done differently, I am open to suggestions.

edwardspec:
Should be good as it is.

Issue #188, opened 2018-12-28
Gold dropped at end of scenario lost


dalong76:
If the final leader that triggers the end of a scenario drops gold, the gold goes into the item storage, but can never come out.

Dugy:
You cannot just take it out of the item storage and walk over it?

Dugy:
Gold isn't listed as an item type when the items are sorted into types, so the gold is retrievable only if there is less than 20 items in the storage.

Shall we fix this now or with the rewrite of the inventory system?

edwardspec:
Rather than showing gold in storage, let's make sure it never gets added to storage.
INSERT_INTO_ITEMS macro can call loti.storage.add(), which can check if the item is gold (and handle it accordingly).

Dugy:
That could be a good approach.

edwardspec:
Resolved by #205.

PR #205, opened 2019-01-02
Avoid gold in item storage

edwardspec:
Possible solution for issue #188.


edwardspec:
Question: do we know how many third-party addons use LotI as a dependency? (if any)

If any of them use INSERT_INTO_ITEMS, we might keep the ARRAY parameter (even if it's no longer used) to not break them.

Dugy:
Add-ons that use LotI as a dependency are _The Beautiful Child,_ _Affably Evil_ and _Kill the King._ I don't think any of them uses that macro, but I am not completely certain.

edwardspec:
They don't use this macro. (just checked)

Dugy:
No problem then.

PR #214, opened 2019-01-03
clear variables and tab to space

dalong76:
I noticed these variables still in my later saved game files and traced them back to here. Also did some tab to space cleanup.
I don't see any flaw here, but I haven't gone more than one scenario ahead.

### DIFF
--- a/scenarios6/15_Apologies.cfg
+++ b/scenarios6/15_Apologies.cfg
@@ -56,6 +56,7 @@
                     image="wesnoth-icon.png"
                     message= _ "He took the ash and transported himself and his allies back into Irdya, the world from which he came. He took a single piece of the ash, put it on his chest, and incinerated it. The effect was incredible."
                 [/message]
+                # Convert active leader (Efraim) into a god
                 [store_unit]
                     variable=Efrstore
                     [filter]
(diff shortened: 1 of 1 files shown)


Dugy:
Thanks for locating that. These memory leaks bloat save files.

Please don't bother with converting tabs to spaces, I am considering changing all spaces to tabs because the thing created for indenting is tab and most text editors allow indenting whole blocks by highlighting them and hitting tab, while doing the same with spaces is impossible. Indenting manually with spaces is laborous. Even the internal representation of WML inside save files uses tabs. Doing anything with space-based indentation is terrible without an automatic tool integrated into an IDE, which is something few people have. I really hate the person who decided that WML files are supposed to be written with spaces and I regret the moment I accepted the pull request that auto-indented all LotI code.

The only reason why I haven't already converted it is that I suck at Python (the tool to indent WML is a Python script) and that it would greatly increase the size of the repository.

matsjoyce:
Which editor are you using? I edit WML using Kate and it can handle spaces just like tabs. Gedit can too IIRC.

Dugy:
I am using both Kate and Gedit, depending on situation, but neither seems to handle spaces. How do you do it?

edwardspec:
In Kate preferences:
kate-options

Dugy:
Oh, thanks.

But here comes the problem that lua is indented with tabs...

edwardspec:
Yes, it's rather annoying to switch this back and forth when editing Lua and WML files in quick succession.

PR #225, opened 2019-01-07
Fix for issue 224

arobinson:
Fix for https://github.com/Dugy/Legend_of_the_Invincibles/issues/224 (also includes deleting some trailing whitespace due to my editor settings)


arobinson:
This fixes the issue, but it may be breaking the item descriptions, so perhaps this is not the right fix
screen shot 2019-01-06 at 6 13 08 pm

PR #239, opened 2019-01-23
clarify melee/ranged bonus applies only to weapons

dalong76:
Current wording is "ranged attacks only", which would mean it also benefits spells. Writing "melee/ranged weapons only" is more explicit.


Dugy:
Yeah, much better.

PR #242, opened 2019-01-28
fix removal of GEARED

### LINKED ISSUE #241: GEARED trait not removed
Obviously I have a history of screw ups, but I cannot get units to lose the GEARED trait if they remove all items. And I fixed it.

dalong76:
fix #241


edwardspec:
Works correctly, merged.
Thank you.

PR #244, opened 2019-01-28
bug fix to make enemy leaders black souls

dalong76:
Apparently it doesn't work in the prestart event but does work in the start event.

I wonder if I'm the only person that ever gets to the end ...


matsjoyce:
I got to the end, but there have been a lot of changes since then.

PR #245, opened 2019-01-29
Make cataclysmal Demon Lords use their firestorm

dalong76:
The AI picks the highest damage, and doesn't know about the storm special. Normal fireball damage is 123, 156, 195, which is higher than firestorm. That means it always picks normal fireball and never picks firestorm. Upgrade firestorm damage so it will get used.

### DIFF
--- a/units/demons_race.cfg
+++ b/units/demons_race.cfg
@@ -552,8 +552,8 @@ shrine=Shrine|Temple|Cathedral|Dome|Cradle
                 [/dummy]
             [/specials]
             range=ranged
-            {QUANTITY damage 120 140 160}
-            number=1
+            {QUANTITY damage 70 90 110}
+            number=2
             icon=attacks/firestorm.png
             defense_weight=0
         [/effect]


dalong76:
Side note: Shouldn't the specially coded version of storm be replaced by {WEAPON_SPECIAL_STORM} ?

Dugy:
It's probably some forgotten legacy code, abilities now check who is whose enemy instead of using lists of enemy sides and there's no problem with enemies having special attacks originally designed to be used by side 1.

dalong76:
I believe attack_weight doesn't actually work correctly, except for zero vs non-zero. I know I tried messing with defense_weight / attack_weight for various things a long time ago (over a year ago with 1.12) and it didn't work properly. Maybe see Wesnoth github issue 1360. When I try just now, it doesn't work. Demon Lord with 120-1 firestorm, but attack_weight=2, (so 240) uses fireball instead (195).

Dugy:
I think it is a value that can be in range between 1 and 0 and some time ago, all values between 0 and 1 were rounded, so it could only be used to disable attacks. However, I don't know how is it now.

Dugy:
Could be done, I think.

Issue #204, opened 2018-12-31
Common set of functions to manipulate unit inventories and advancements needed


Dugy:
All modifications of units are currrently stored in units' modifications arrays. However, they are used quite rarely, while the whole array is copied often. It would be useful not to store them at all and store only the ids of advancements and numbers of items (in some cases, also sorts). All other information can be obtained elsewhere.

This transition would be quite wide and the code is not ready for it. To prepare for the transition and make it as fluid as possible, it will be necessary to isolate all code that accesses the items.

The functions should accept either a WML array or unit's id as their first arguments and use the variable type to decide what to do with the first argument. The other arguments would depend on the function. The plan is to make them all accessed by ids, but the WML tables might be needed with the current implementation.

Further thought will be needed regarding the specific functions.

matsjoyce:
This sound like a good idea. It would reduce save size and make changes we make appear in the next scenario. Currently changing an item requires next scenario + unequip + equip, which makes savefile hacking a bit more difficult.

edwardspec:
I will provide a sketch on the API and underlying architecture of this.

The migration idea is to implement this API twice:
1) one will work via the new efficient Lua array-based storage (with indexes and all),
2) one will work the old-fashioned way, via user.modifications[].

Then all new code can use the API, even though a lot of WML still users user.modifications[] directly.
Later, after all the old WML is rewritten to use the API too, we can use implementation (1) as default, with (2) as fallback for old saves.
Even later, (2) is removed.

edwardspec:
Oversimplified view is that we have an array

```lua
A = [ 'user_id1' => [ data1, data2, data3 ], 'user_id2' => ...]
```

... where `data` is an arbitrary Lua array (not only the items can be stored there).

There will be search methods by certain fields in data[],
but they won't just loop through the entire A array,
they will use indexes that I will implement (caller of the API doesn't need to know about them).

edwardspec:
If `data` is an item, then it probably just needs `number`, `sort` (because crafted items can be of different sort with the same number) and `description` keys (because description is slightly different due to locked/unlocked latent bonuses, which are coloured differently). The rest is the same as `loti.item.type[number].something`.

matsjoyce:
The description is usually generated on the fly isn't it?

edwardspec:
> The description is usually generated on the fly isn't it?

Currently it's calculated in `add_set_effects()` in `stats.lua` and saved in `user.modifications`.

[...]

PR #250, opened 2019-02-05
Merge unitdata library into master branch, step 1: unitdata.lua + testsuite

edwardspec:
First step of merging `unitdata.lua` into master branch.
See https://github.com/Dugy/Legend_of_the_Invincibles/issues/204#issuecomment-457998180 for details on step 1 and step 2.

Note: the library is NOT used anywhere yet. All changes in `stats.lua` and `items.lua` (which actually affect the game) were **excluded** from this branch (`unitdata-api-lib`). They can be found on `unitdata-api` branch.
They will be merged separately (as step 2 of the merge).


Dugy:
I know you're still doing something around it, but I have some remarks:

- Please don't use points at the end of comments as long as the comments aren't several sentences long, it's not prose
- Why was the listing of AMLA effects removed from the `effects` function? My implementation of `stats.lua` depended on it, although it seems not to be merged into master yet. Why is that?

edwardspec:
> Why was the listing of AMLA effects removed from the `effects` function?

They are still listed.
I replaced three loops (for advancements, items and other modifications) in `effects()` with one loop around all modifications. This covers the advancements too.

The test unlocks "Better with the sword" and "Harder to hit" on Efraim (Demigod) and he correctly gets the effects:
https://github.com/Dugy/Legend_of_the_Invincibles/blob/34404dc30adc45e9f5018cc9f8a4293224ed0dcf/tests/lua/unitdata.test.lua#L373
https://github.com/Dugy/Legend_of_the_Invincibles/blob/34404dc30adc45e9f5018cc9f8a4293224ed0dcf/tests/lua/unitdata.test.lua#L377
https://github.com/Dugy/Legend_of_the_Invincibles/blob/34404dc30adc45e9f5018cc9f8a4293224ed0dcf/tests/lua/unitdata.test.lua#L443-L472

Dugy:
I was wrong.

But I still can't find where it lists traits.

edwardspec:
Traits are modifications too.
This is also covered by the test.

Dugy:
I see how it's done now. But why did you overwrite my code when my code was written in a way that was specifically prepared to be easily adaptable to the removal of equippable items and selectable advancements from the units' codes?

edwardspec:
I rewrote it because its code is 3 times smaller now.
The implementation based on Lua arrays won't use this function at all. This function is only for WML-based storage.

Dugy:
The lua implementation will be very similar to my code. I didn't want to write it twice. You didn't have to do it, but certainly, the code looks better now.

edwardspec:
I'll provide the Lua implementation. (after stats.lua is tested and merged)

For the time being, we need both WML storage code and Lua storage code at the same time (we shouldn't overwrite the WML code), so that when someone loads an old save file, it can be read by WML storage code. (and converted into Lua storage format)

edwardspec:
> Having the lua functions fall back to the WML implementation may be more difficult

```lua
if is_this_an_old_save() then
	loti.unit = wml_based_implementation
else
	loti.unit = lua_based_implementation
end
```

Here `is_this_an_old_save()` can, for example, check `wml_based_implementation.items()` of one unit with GEARED trait. If it returns anything, then this is an old save.

> Or you mean to write some code that will actually convert everything to the lua storage when old save file content is detected?

Can do this too. When the save file is loaded, we can check "is this an old save?" and then move items/advancements into the Lua storage. That's even better.

edwardspec:
> I think you can just take my code and change a few lines

Just have a look at this **very preliminary** sketch.
Specifically at the code of `items()`, `advancements()` and `effects()`.

(all the heavy lifting is moved from `effects()` into utility functions `nextval()` and `effects_iter()`, which do trivial, easy-to-test things)

```lua
-- Sample storage (1 unit)
storage = {}
storage.items = {
	{ number = 100, effects = { "Effect 1", "Effect 2" } },
	{ number = 200, effects = { "Effect 3", "Effect 4" } },
	{ number = 300, effects = { "Effect 5", "Effect 6" } }
}
storage.advancements = {
	{ id = "fireball", effects = { "Effect 7", "Effect 8" } }
}
storage.other = {
	{ tag = "trait", name = "strong", effects = { "Effect 9", "Effect 10" } }
}

-- Utility function. Returns stateless version of next() for { number => value } array.
-- Example:
--	f = nextval({ 'a', 'b', 'c' })
--	print(f()) -- Prints "1 a"
--	print(f()) -- Prints "2 b"
local function nextval(array)
	local idx = 0
	return function()
		idx = idx + 1
		if array[idx] then
			return idx, array[idx]
		end
	end
end

function items() return ipairs(storage.items) end
function advancements() return ipairs(storage.advancements) end
function other() return ipairs(storage.other) end

-- Iterator over some of the effects.
-- Parameter: next_modification() - function that returns the next modification.
local function effects_iter(next_modification)
	local idx = 0
	local next_effect = function() end

	return function()
		while true do
			local _, effect = next_effect()
			if effect then
				idx = idx + 1
				return idx, effect -- Found another effect
			end

			-- Try next modification
			local _, modif = next_modification()
			if not modif then
				return -- Nothing more to return
			end

			next_effect = nextval(modif.effects)
		end
	end
end

function effects()
	local iterators = {
		effects_iter(nextval(storage.items)),
		effects_iter(nextval(storage.advancements)),
		effects_iter(nextval(storage.other))
	}

	local idx = 0
	return function()
		for _, candidate in ipairs(iterators) do
			local _, val = candidate()
			if val then
				idx = idx + 1
				return idx, val
			end
		end
	end
end

for idx, eff in effects() do
	print("Found effect #" .. idx .. ": " .. tostring(eff))
end
```

Dugy:
Okay.

PR #254, opened 2019-02-07
Merge unitdata into master branch, step 2: stats.lua + secondary bugfixes

edwardspec:
See also: issue #204 (discussion of unitdata API), pull request #250 (where unitdata library was added).

### DIFF
--- a/lua/items.lua
+++ b/lua/items.lua
@@ -486,11 +486,7 @@ loti.item.describe_item = function(number, sort, set_items)
 	local desc = {}
 
 	if item.defence then
-		local defence = item.defence
-		if item.sort == "armourword" and (sort == "boots" or sort == "helm" or sort == "gauntlets") then
-			defence = defence / 3
-		end
-		defence = math.floor(defence)
+		local defence = math.floor(item.defence)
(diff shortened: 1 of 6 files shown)


Dugy:
You can merge it if you think it's time.

edwardspec:
Looks ok to me, merging.

Dugy:
Regarding the transition, I think it's currently more important to transfer any inventory interaction to the new API than to implement the new behaviour.

Not only because the new implementation will break all code not using the API, but I would also like to make one 'stable' release before it so that there would not be any WML left in save files that accesses the data the old way while the lua is up-to-date.

edwardspec:
> I would also like to make one 'stable' release

Current state looks ok for a stable release (unless any of the bugs reported on forums are blocking it), you can just release it in a few days from now.

We should have at least 1 release version before changing the WML, as to not introduce any last-minute bugs to the release.

Dugy:
I don't think we've done enough visible changes to justify a new version at the moment. I'd prefer to have everything in WML converted, then wait for a while and release it then.

PR #263, opened 2019-11-29
Fixed damage calculation in bonus attacks.

Chergnobarp:
Wesnoth uses integer values for damage and attacks, while lua uses floats. So Guide to Lycanthropy was choosing a 18-5 (damage = 90) attack over a 16-6 (damage = 96) attack, because it was actually seeing 16.85-6 (damage = 101.1) and 18.5-5.75 (damage = 106.375).


Dugy:
Thank you.

PR #268, opened 2020-01-09
Fix to allies

dalong76:
Allies were not on same team (capitalization appears to matter).
Allies also moved, so now they are set as idle. I'm not sure why they didn't move in the past though.


Dugy:
Looks like a typo to me. Thanks for fixing it.

PR #270, opened 2020-04-21
Can now purify Zombie Riders

Chergnobarp:
For future-proofing the code, it might be worth rewriting it as:

```
    if (unit is undead) {
        if (unit is an archer) {
            archer units here
        } else if (unit is mounted) {
            mounted units here
        } else {
             melee units here
        }
    }
```

Better to create the wrong unit type than to create none at all.


Dugy:
The game has no track of what units are archers and what units are mounted, so there's unfortunately no good way to implement the future-proofing you've suggested.

Chergnobarp:
I meant that as a shorthand for what you're doing now - if unit is bone shooter or skeleton archer etc. Now that you mention it though, I wonder if the values for usage/recruitment_pattern could be used here.

Dugy:
That is a property of the unit type, not of the unit, but in theory, it could be looked up. It's also somewhat inaccurate, but it could be workable. I will open a feature request for that.

PR #272, opened 2020-04-24
UI improvement for Investigations in ch8

Chergnobarp:
I've never liked the way labels appeared in this scenario. They often appear offscreen, so the first time player doesn't even realize that a label's been placed. Even if they do, tracking it down is tedious. With this change, the screen is temporarily scrolled to the goal location and a marker is placed, so the player knows where to go.

I could fold the objectives updates into this too, but I figured I'd see what you thought before I did any more work.


Dugy:
I have no complaints about your work. I can only thank you for contributing.

PR #273, opened 2020-04-24
Now collecting items from impassable terrain at the end of the scenario.

Chergnobarp:
This covers Beelzebub in a few scenarios, and units that spawn in the walls because the trapdoor/portal hex was occupied.


Dugy:
Thanks.

Issue #233, opened 2019-01-19
Titles given to inappropriate units


Dugy:
Units that shouldn't get titles:

* Low level units that don't advance
* Infernal plants
* Infernal machines
* Imps

Anything else?

edwardspec:
Evil Lethalia, resurrected Argan in Chapter 7, maybe other key figures like Niflheim.
They are famous and their names are unique, adding titles seems unnecessary.

Dugy:
Units that are set to be unrenamable don't receive titles. Of course, there are instances where I forgot to mark them as unrenamable.

PR #275, opened 2020-04-26
More selective about giving titles to units.

Chergnobarp:
Fixes https://github.com/Dugy/Legend_of_the_Invincibles/issues/233.

The gist of this is to remove titles from units without advancements. That covers everything in the linked issue, plus merchants, flies, spell eaters, etc, etc. I'm open to changing the logic, but I think that's pretty close to how it should be. The only exception I'd make is to give titles to the officers of spawned parties, e.g. infiltrated celestial messengers in Jungle Hell and corrupted elvish juggernauts in Arctic Wasteland. I think the best way to do that is to give them hero icons.

I made sure that unadvanceable nameless units still get names without the titles.

There seems to be a mix of tab and space indentation in this file. I'm a tab man, so I used tabs for this.

### DIFF
--- a/lua/main.lua
+++ b/lua/main.lua
@@ -744,10 +744,20 @@ function wesnoth.wml_actions.check_unit_title(cfg)
 		return
 	end
 	unit_variables.been_given_title = true
+
+	-- If the unit has no name, give it one
+	if u.name == "" then
+		if not u.advances_to or u.advances_to == "null" or u.advances_to == "" then
+			if u.race == "undead" then
+				u.name = undead_names()
+			else
+				u.name = nameless_generator()
+			end
+		end
+	end
+
 	local deserves = false
-	if not u.advances_to or u.advances_to == "null" or u.advances_to == "" then
-		deserves = true
-	elseif u.canrecruit then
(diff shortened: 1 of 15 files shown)


Dugy:
Thanks.

It doesn't fix the part of the issue that the villain Efraim or villain Lethalia in chapter 6 gets the title. They only need to be set to be unrenamable, but I never remember I should do it. Will you do it or will you leave it for me until I finish something and get to it?

I also prefer tabs (because indentation is what they are meant for and they grant the freedom to choose how wide they are without altering the project), but the automatic WML tools enforce spaces. So I use tabs in lua and leave the WML automatically formatted to spaces, but use git filter to use tabs locally. See README.md to for instructions how to set it up.

Chergnobarp:
Considering that this has been a known issue for more than a year, I'd better do it :)  It's not clear to me exactly which units count as famous, but this should be a good start.

Also, shouldn't Abaddon's race be demon lord instead of demon?

Dugy:
Thanks.

Yeah, Abaddon's race should better be a demon lord.

PR #276, opened 2020-04-28
Extra damage from high-level hybrids of steel is user-visible.

Chergnobarp:
The first time I played chapter 5, when Akula revealed her true form, I attacked her with Efraim's lesser berserk. Imagine my surprise when he started taking significant damage for no discernible reason, and eventually died from it. I made this into an ability, so now the player doesn't get any nasty surprises.

Akula's extra damage was pierce before this change.

### DIFF
--- a/units/Akula_steel.cfg
+++ b/units/Akula_steel.cfg
@@ -84,6 +84,7 @@
     #      [/portrait]
     {TRAIT_MECHANICAL}
     [abilities]
+        {ABILITY_TENTACLES}
         {ABILITY_REGENERATES}
     [/abilities]
     [attack]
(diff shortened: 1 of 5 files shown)


Dugy:
Good idea.

There are a few errors, though. Events in units need to have unique id= keys, otherwise they may be duplicated. This could cause Akula to deal the tentacle damage many times after being attacked.

Another issue is that the tentacles are supposed to do variable damage in the final area, depending on the number of Amplificators you have destroyed. Your changes do not seem to support that.

Chergnobarp:
For your first point, I removed the events in units. There's nothing to duplicate. I tested Akula in the walking in shadows scenario, it works properly for me. I just copied what you did for thorns, which also doesn't have event ids.

For the second point, nothing in the final battle has the new tentacles ability (there's a separate unit for Akula in that scenario). The amplificator-based events still stand unchanged, and work as always. Well, the turncoat from the prison might have it, but the amplificator-based events don't apply to side 1, so that's a non-issue.

Dugy:
Okay, better now. But the Akula from the final scenario doesn't have the ability despite being possibly more deadly.

Chergnobarp:
Good point. I just added some extra abilities for that scenario.

Dugy:
I don't think it can work that way. If you want to add an `[effect]`, use the `[object]` tag.

And you've forgotten to add a space between the ability name and the damage in brackets.

Chergnobarp:
I know it can work that way ([as of 1.13.6](https://wiki.wesnoth.org/DirectActionsWML#.5Bmodify_unit.5D)) because I tested it before committing. I even edited my save file to get different values for amplificators_destroyed. Seems that neither of us noticed the missing `[effect]` tag for Akula. Not sure how that worked, but it did.

Regardless, they're both using `[object]` now, and I changed the paren spacing too.

Dugy:
And are you sure it works the way it is? I am quite sure that `[object]` doesn't work within `[modify_unit]`, it can be a standalone tag with its own `[filter]`. And she may have a special ability given her unit definition through `[modifications]` as in [here](https://github.com/Dugy/Legend_of_the_Invincibles/blob/master/scenarios1/15_Long_Way_Home.cfg#L96).

Toranks:
Maybe it's too late to comment because surely things will have changed since then on this files, but modify_unit can use  [object] inside, and in fact it is usual and convenient to do it, although there are other methods.

Dugy:
The branch has been merged already, so it doesn't really matter anymore.

PR #278, opened 2020-04-30
Added editorconfig.

Chergnobarp:
[EditorConfig](https://editorconfig.org) helps automatically set the right indentation settings in editors. Supported by many out of the box, and there are plugins for the rest.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+root = true
+
+[*.{wml,pbl}]
+indent_style = space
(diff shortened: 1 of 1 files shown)


Dugy:
I am using git filter to have tabs also in WML (with width 8), so I specifically do _not_ want to have those settings. My editor doesn't support it without plugins, but I might decide to use some other editor at some point.

I will add a note how to ignore it locally.

PR #279, opened 2020-05-06
Allow all three Tome of Bards Volume II amlas

dalong76:
With Tome of Bards Volume II, two of the three AMLAs have the same id, which seems to make them mutually exclusive. I don't think that is intentional.

### DIFF
--- a/utils/amla.cfg
+++ b/utils/amla.cfg
@@ -1320,7 +1320,7 @@
     [/advancement]
     [advancement]
         max_times=1
-        id=ToBVII2
+        id=ToBVII3
         strict_amla=yes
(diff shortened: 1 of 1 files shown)


Dugy:
It _was_ intentional. Do you think it would be better to make it non-exclusive?

dalong76:
The flaw is it isn't obvious you can only pick one. I noticed because I tried to pick them all and couldn't.  

Relatively, I don't think the AMLAs have similar value. The first one (20% resistance leadership) is awesome. The other ones are not nearly as great.

Dugy:
So there are two options. I can erase make it clearer they are mutually exclusive (and balance it) or I can accept your pull request.

Because you are a longtime player, I am giving you the choice.

PR #288, opened 2020-08-05
Efraim's name really made translatable

Discontinuum:
Sorry for the second pull just in a few minutes, Efraim in the side tag is probably only a notion for testing, the real Efraim is created here when you play the campaign, so it's here his name must be translatable.


Dugy:
It's better to leave it the way it is for consistency.

Discontinuum:
It won't be shown translatable in game without this addition I fear.

Dugy:
Seems I have read the changes wrongly. Maybe I should not check pull requests at 1 a.m.

PR #308, opened 2021-06-06
Make possible unequipping items

Discontinuum:
The latest fix to disallow unequipping limited items broke unequipping of any item because of missing parentheses


Dugy:
I program in C++ every day for many years now. Lua has similar operator priority to C++. I don't understand how could I get this wrong.

Discontinuum:
Folks in LotI's forum topic point out that `item_sort ~= "limited"` was probably even better than not and parentheses around equality, though it works either way

Issue #302, opened 2020-12-10
For some units an equipped weapon doesn't apply


Discontinuum:
For some units a weapon they're allowed to equip doesn't apply (in a sense that it doesn't change the attack's name and stats, resistances and moves and other non-attack properties work) when they actually equip it. Though logically it shall apply. A list of some examples (most likely incomplete):

Drake Fighter and Drake Warrior can equip a sword, but it doesn't apply. The attack's name is "war blade".

Drake Blademaster can equip a sword, but it doesn't apply. The attack's name is "war talon".

Dune Rider can equip a bow, but it doesn't apply. The attack's name is "composite bow".

Al Ghareeb from "The Beautiful Child" campaign can equip a sword but it doesn't apply. The attack's name is "long sword".

I suppose this might happen because all these examples above have slightly "non-standard" attack's name, not just "sword" or "bow". So the game is able to recognize a correct item type but for some reason it doesn't recognize it when calculates attacking stats

dalong76:
Where can you get a drake unit on your own side? What is a Dune Rider?

Discontinuum:
Dune Rider is https://units.wesnoth.org/1.14/mainline/en_US/Dune%20Rider.html

Well, Drakes and others I got in my experimental add-on where I plugged LotI items and drops to World Conquest II. But even omitting my add-on, the bug exists and has its effect in another dependent on LotI campaign, The Beautiful Child. You can look at this save ([TBC-Ziggurat Town-Auto-Save1.gz](https://github.com/Dugy/Legend_of_the_Invincibles/files/5675821/TBC-Ziggurat.Town-Auto-Save1.gz)), you'll need "The Beautiful Child" add-on to be installed.

Discontinuum:
More on this, the code in lua/stats.lua recognizes as sword attacks only a fixed list of attacks' names
`if wn == "sword" or wn == "short sword" or wn == "greatsword" or wn == "battlesword" or wn == "saber" or wn == "mberserk" or wn == "whirlwind" or wn == "spectral blades" or wn == "lichsword" then
			weapon_type = "sword"`

While in lua/main.lua (the function which determines a list of allowed item types) any unit with any attack that satisfies "sword$" regex is allowed to equip a sword. A similar story is with bow. With "war talon" and "war blade" the things are more interesting: in stats.lua they are classified as some "exotic" while in main.lua they simply go as "sword".

Maybe this mismatch is the cause of the problem.

[...]

Issue #300, opened 2020-12-06
Known issues with multiplayer


Discontinuum:
1. https://github.com/Dugy/Legend_of_the_Invincibles/issues/298
2. Other after-crafting options (smashing, drop on the ground) also cause OoS
3. "Pick up from the ground" causes OoS: other players don't see the item picked, for them it's still on the ground
4. Enemy units cause item events (picking a gem, gold, an item, though gems go to the players' storage and items' fate decided by an mp host [he got an item window])

edwardspec:
_(disclaimer: I don't have capability to test the game in multiplayer, so I can't fix this myself, but the information below should be useful for whoever tries to fix it)_

By its design, the Inventory dialog itself is resistant to OoS errors, because it does every item-related operation via
```lua
inventory_dialog.mpsafety:queue({ ... })
```
... rather than invoking them directly.

The catch is, **any item-affecting changes** that happen **while the player has the Inventory dialog open** should also be performed via `mpsafety:queue()`. Otherwise in multiplayer these changes will be performed in another order for other players.

Both Crafting and "Pick up from the ground" are calling an event "item_pick", and that event is not using `mpsafety.queue()`. This is what causes OoS.

Simplest solution is probably to explose the variable `inventory_dialog.mpsafety` (make it not `local`) to be usable from outside the Inventory dialog's code (and use it in event handler of `item_pick`).

Discontinuum:
I have found the reason for the 4th issue: enemies picking items in MP. The trick is that an item dropped from a defeated enemy is added with the "loti.item.on_the_ground.add" facility from lua/items.lua. And when it adds an event handler for picking the item, it uses filter_side "controller=human". But according to https://wiki.wesnoth.org/StandardSideFilter , this filter cannot be used in multiplayer games. That's why in MP enemies pick items while in SP everything's fine

Discontinuum:
Found
5. Gems transmuting doesn't work, other players don't see changes

Discontinuum:
[...]

PR #309, opened 2021-06-07
Weapon bindings system

Discontinuum:
As mentioned in #302 we need a unified table of all weapon bindings, so I suggest my solution.

Not exactly a WML resource table, like items' list. I made a special tag `add_weapon_binding` which adds a "name->type" binding to a table `loti.item.weapon_bindings`, then both stats.lua and main.lua use that lua table to determine items' effects and allowed items to gear. Additional bindings are easy to add for other addons using LotI. Also, as mentioned in the issue, I made Drakes' war talon and war blade to be metal claws instead of sword/exotic. Also added cleaver as an axe (in the previous system cleaver has a reversed bug: axe would work for it if equipped but the system doesn't allow to equip it.

My only concern about this solution might be the compatibility with other add-ons using LotI, if they don't use LotI's GLOBAL_EVENTS macro (which now loads DEFAULT_WEAPON_BINDINGS). I checked other Dugi's campaigns, and they all use it, and my addons I will easily patch when the new major LotI release comes

### DIFF
--- a/lua/main.lua
+++ b/lua/main.lua
@@ -841,47 +841,8 @@ function loti.util.list_equippable_sorts(unit)
 	-- Analyze the list of attacks. Allow weapons that are logical for this unit.
 	for attack in pairs(loti.util.list_attacks(unit)) do
 		local weapon_type = loti.item.weapon_bindings[attack]
-		if weapon_type == "sword"
-				then can_equip.sword = 1
-		elseif weapon_type == "axe"
-			then can_equip.axe = 1
-
-		elseif weapon_type == "staff"
-			then can_equip.staff = 1
-
-		elseif weapon_type == "xbow"
-			then can_equip.xbow = 1
-
-		elseif weapon_type == "bow"
-			then can_equip.bow = 1
-
-		elseif weapon_type == "dagger"
-			then can_equip.dagger = 1
-
-		elseif weapon_type == "knife"
-			then can_equip.knife = 1
-
-		elseif weapon_type == "mace"
-				then can_equip.mace = 1
-
-		elseif weapon_type == "polearm"
-				then can_equip.polearm = 1
-
-		elseif weapon_type == "claws"
-			then can_equip.claws = 1
-
-		elseif weapon_type == "sling"
-			then can_equip.sling = 1
-
-		elseif weapon_type == "essence"
-				then can_equip.essence = 1
-
-		elseif weapon_type == "thunderstick"
-			then can_equip.thunderstick = 1
-
-
-		elseif weapon_type == "spear"
-				then can_equip.spear = 1
+		if weapon_type then
+			can_equip[weapon_type] = 1
 		end
 	end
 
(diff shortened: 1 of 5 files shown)


lua/main.lua
```diff
@@ -840,60 +840,47 @@ function loti.util.list_equippable_sorts(unit)
...
 			then can_equip.thunderstick = 1
 
 
-		elseif attack == "spear" or attack == "javelin"
-			or attack == "lance" or attack == "spike"
-			or attack == "pike" or attack == "trident"
-			or attack == "trident-blade" or attack == "pitchfork"
+		elseif weapon_type == "spear"
 				then can_equip.spear = 1
 		end
```
Dugy, line 885: Why not just replace this whole thing with `can_equip[weapon_type] = 1`?

Dugy (commented):
I think you have kinda overcomplicated it.

Dugy:
Where do my other add-ons use this? I can't think of anything that could cause compatibility problems.

lua/main.lua
Discontinuum: Right, I have forgotten that dot and brackets notation are of the same meaning

Discontinuum:
> Where do my other add-ons use this? I can't think of anything that could cause compatibility problems.

Your addons use standard LotI's {GLOBAL_EVENTS}, so they should run fine. My MP addons use their own patched {GLOBAL_EVENTS} instead of LotI's to workaround some MP bugs

Dugy:
> Your addons use standard LotI's {GLOBAL_EVENTS}, so they should run fine. My MP addons use their own patched {GLOBAL_EVENTS} instead of LotI's to workaround some MP bugs

Aha. Are these bugs in LotI?

Discontinuum:
> Aha. Are these bugs in LotI?
Well, I use custom GLOBAL_EVENTS for several reasons. One is that original LotI code always considers 1,2,3,4,5 sides to be players, though I'm working with WC, I have only 1,2,3 to be players. Also I had to fix some custom advancements options (one of the fixed, for Elvish Lord, I uploaded here, but it's not in the stable so I had to patch it in the addons).
Also some minor things like autopicking items on the last turn didn't work (victory isn't triggered for some reason, enemies defeated is), and a patched update_stats (it erases roles of the unit, for LotI it's insignificant, but WC II uses unit roles).

Dugy:
I think the side 1,2,3,4,5 could be replaced by something smarter, such as filtering for sides with human controllers. I can merge the role fix into LotI if you want.

Discontinuum:
> I think the side 1,2,3,4,5 could be replaced by something smarter, such as filtering for sides with human controllers. I can merge the role fix into LotI if you want.

With checking for AI - I fear that can't work as straight-forward as it may seem. One of the issues with items, AI picking items in MP, is caused by the fact that the used checking facility for AI is OOS-unsafe and doesn't work in multiplayer. I mentioned it here https://github.com/Dugy/Legend_of_the_Invincibles/issues/300#issuecomment-751111248

I thought about something like a variable with friendly sides could work though

Dugy:
Hm, then it would have to be a variable, whose value would have to be determined somehow. Not nice.

PR #311, opened 2021-06-08
Correct MP extra advancements for loyalist cavalry and orcish warlord

Discontinuum:


Discontinuum:
Don't know why two previous commits also go in there to be honest

Discontinuum:
Also I can't really find any place where a unit with "Grand Knight LotI" type was used

Dugy:
That was a mistake, there is no such unit as _Grand Knight LotI._

Issue #313, opened 2021-06-09
Critical bug with weapon bindings


Discontinuum:
They are written into the lua table which isn't saved and dropped when a user reloads a save.

You said to me in the PM that you use dummy unit types, do you mean these https://github.com/Dugy/Legend_of_the_Invincibles/blob/master/units/data_loaders.cfg ? So the only things to load in lua to prepare things before the commit have been these titles and item list? I thought item list is prepared in global events where I actually put my weapon bindings tags macro

Dugy:
It used to be that way, but when the core was rewritten to lua, it was changed. This has the consequence that the list of items is not saved in the save file, making it smaller. Also, changes to items take effect after reloading the game, the scenario doesn't have to be started from the beginning.

You can make your own data loader pseudounit for those weapon bindings, but you'd get a WML table which is quite annoying to work with in lua. Moving that into a preload event would be easier.

Discontinuum:
Seems fine now

PR #314, opened 2021-06-10
Make weapon bindings be reinitialized using preload

Discontinuum:
Fix for #313 . Old midgame saves which lack this preload events will still be broken, but only till the end of the scenario.


Dugy:
Thanks.

PR #316, opened 2021-06-11
Synchronized soul eater/redeem

Discontinuum:
The soul eater/redeem menu has been unsynchronized which led to negative effects both for MP and watching SP replays. In MP it showed the redeem menu for each player and they all had to choose the same option (or they would get a nasty oos some later), in SP replays it's kinda the same, if your character got a new advancement path, the menu pops up and you have to choose the same option as you had in-game, the choice wasn't remembered in the replay. This should fix that, although old unsynched replays won't work after that (they will show oos "input expected but found something else"


Dugy:
Thanks.

Issue #304, opened 2021-02-06
Vritra just disappears if you use the dark blast too much


Discontinuum:
Vritra is supposed to turn into a more corrupt version of herself when she uses the corruption attack too much, but it seems it works incorrectly and instead of changing the variation she just disappears. There's a replay for that
[LotI2-Overpowered_replay.gz](https://github.com/Dugy/Legend_of_the_Invincibles/files/5937036/LotI2-Overpowered_replay.gz)

Discontinuum:
This code https://github.com/Dugy/Legend_of_the_Invincibles/blob/c73371eeada7a0fa962dfdbc2740ac59303abe45/units/Vritra.cfg#L891 causes "failed to auto-store $unit at (44,9)" (coords of the unit) printed to console and she disappears

Dugy:
It looks like old code from before the time when update_stats was turned into a function. It was created in a variable that was then used in the updating code, however, that code does not exist any more and the unit is expected to exist. I think that simply removing the `to_variable` line and replacing the macro by a proper use of the `[update_stats]` tag should fix it.

PR #319, opened 2021-06-12
Fix Vritra's corruption logic

Discontinuum:
Fix for #304, tried it till the first variation, seems working (she blinks during a transition but that's okay compared to an incorrect disappearance)


Dugy:
I think the blinking was there even before it was broken.

PR #321, opened 2021-06-14
Fix non-working lookup of weapon specials for titles

Discontinuum:
The previous code didn't actually looked up the specials, it reached the tag `[specials]` and queried its `id` (which is absent ofc), for actual looking up you need to iterate over `[specials]` children


Dugy:
Thanks.

Issue #320, opened 2021-06-14
 Make "gem found" window non-blocking (at least in MP)


Discontinuum:
A blocking gem window is uncomfortable in MP because players need to click it to go through, so a player can't go afk and wait until the turn bell rings to go back to his move, because the gem window will block the game until he clicks.

Dugy:
Are the gem windows different from the item windows? Do you mean that all players have to wait until all players click on the pickup window to continue?

That seems to be a serious problem. That window is mostly cosmetic, finding its source and replicating it in lua inside a synchronised section should fix it. It could also make the code look more readable since the `[object]` tag has little to do with its effect on the unit.

Discontinuum:
Well, not so dire, others don't have to wait everyone to proceed. But still it is somewhat annoying when you need to click through on someone's else turn. When it's your turn, it's okay, but when it's not your turn and you have to click, not nice. Can't go afk and wait for the turn bell without checking, as I mentioned. Also tiresome for observers to click through all gems to go through the replay

Discontinuum:
Not relevant after #323

Issue #298, opened 2020-11-29
Crafting an item and equipping it immediately causes OoS


Dugy:
If an item is crafted and put into the item storage, everything is OK, but equipping it immediately results in an OoS.

edwardspec:
See #300 for details on what causes this.

Dugy:
Yes.

PR #323, opened 2021-06-17
Overhaul of item_pick and multiplayer synchronization

Discontinuum:
Fix for #298 , a good deal of #300 and additionally #320 

I decided to write item_pick event in lua because the WML code contained duplication of what is in lua, and it was quite fragile (like it had to setup fearless for book of courage manually while in lua it's all done by itself applying item's `on_equip` code.

There are some differences in behaviour, now pop ups about items and gems appear only for the current player (and not for others), also money causes pop-up too (so now everyone can read money items' descriptions, not just translators).

I decided to make two version of item_pick event becayse I wasn't sure about correctness of synchronize_event inside another synchronize, and overall about synching events from two mpsafety queues at the same time.


Discontinuum:
Disabling attacks for retaliation is unsynchronized for now

Discontinuum:
Also, `loti.item.storage.add` method now handles gems, not only money, so I guess it might resolve the issues with gems getting into the item storage instead of their special variables

Discontinuum:
`can_equip_item` tag probably should be returned to save compatibility with older saves/replays, though it's not used in the current code anymore

Dugy:
Thanks for fixing this.

Discontinuum:
Hm, the last update of LotI beta is June 15th, I wonder if everything's okay? I expected it to be updates this night, so I could upload beta versions of my addons (to use LotI's beta instead of stable) and go full-scale testing it with friends

Dugy:
@edwardspec Do you know what could be going on? I am quite clueless about this CI stuff.

edwardspec:
Travis CI has discontinued support for opensource projects, so its checks are no longer running.
`.travis.yml` needs to be rewritten to use GitHub Actions instead.

Dugy:
If that isn't some kind of trivial change, then LotI beta is doomed, since I have no knowledge of this stuff.

edwardspec:
I will look into this.

Dugy:
Great, thanks.

Discontinuum:
It's written that the job run successfully this night but the add-on wasn't updated

edwardspec:
From commit description:

> 3) "upload beta addon" requires PBL_PASSPHRASE and PBL_EMAIL secrets
> to be set in the repository settings:
> https://github.com/Dugy/Legend_of_the_Invincibles/settings/secrets/actions

Dugy:
I have set them, however, I cannot recall what as the password on the server. I cannot even recall setting it to the repository, maybe it wasn't me who set it there. I think that Edward is implying I had set it originally, so I am guessing that it was probably the same as in the main version, as there is no reason to use a different one. I recall some other passwords I occasionally used, so if it was really me who set the password, I'll probably remember the right one, no worries.

edwardspec:
PBL_PASSPHRASE should be set to the same password that was used when the beta was uploaded to Wesnoth addon server for the first time.

From the log of https://github.com/Dugy/Legend_of_the_Invincibles/runs/2910105905 (you can click on the line `Run .github/scripts/upload_beta_addon.sh` to see the relevant part of the log),
addon server says that current value of PBL_PASSPHRASE is not the correct password:
```
Run .github/scripts/upload_beta_addon.sh
('Ignored dir', 'tests')
('Ignored file', 'beta.pbl')
107/107
Opening socket to add-ons.wesnoth.org:15015 for 1.15.x
Connected as 42.
Adding directory Legend_of_the_Invincibles_beta as Legend_of_the_Invincibles_beta.
Receiving 107 bytes.
Received 107 bytes.
Add-on rejected: Incorrect add-on passphrase.
```

edwardspec:
https://forums.wesnoth.org/viewtopic.php?f=15&t=24320
> In the case of a lost password or _server.pbl file:
> Please send a private message to the Forum Administrators group.

Discontinuum:
> The beta seems to have a different password than the main version. I am not sure if I can remember what it was.

I wonder if you could log in in your Travis account and watch it there. You have set it somewhere in the previous system, maybe it's possible to look there?

edwardspec:
> I wonder if you could log in in your Travis account and look for it there.

Not possible (secrets are not shown even to a person who added them). The value is only seen from Travis scripts, which are not running anymore.

Discontinuum:
Till the moment it hasn't worked

PR #326, opened 2021-06-20
Fix of info menu for units with defence cap

Discontinuum:
For #324


Dugy:
Thanks.

If your commit message is `Fix #326`, it will automatically close issue 326 and link the commit in the closing message.

PR #329, opened 2021-06-30
Improve parry bonus attacks

Discontinuum:
I first thought about making it even -50%, but decided to restrain myself a bit and make it 55%. For units with 4 base strikes it makes 1 parry strike, but they got 2 attack after one "parrying better" advancements. For units with 5 base strikes it gives 2 parry strikes. So overall parry becomes beneficial against enemies with 3 or more strikes when you're on a good terrain. Although having 1 parry strikes for units with 4 strikes might seem not very convincing for new players (like doesn't look very powerful), after all I decided that having 2 parry strikes for 4 strikes base attack right from the only advancement is too much (and the only unit with 4 base strikes to have a "parry" advancement is two variations of Delly)

### DIFF
--- a/units/Delly_lich.cfg
+++ b/units/Delly_lich.cfg
@@ -828,7 +828,7 @@
                 clone_anim=yes
                 defense_weight=0
                 damage=-20
-                number=-55
+                number=-30
                 [specials]
                     {WEAPON_SPECIAL_PARRY}
                 [/specials]
--- a/units/Delly_lich.cfg
+++ b/units/Delly_lich.cfg
@@ -828,7 +828,7 @@
                 clone_anim=yes
                 defense_weight=0
                 damage=-20
-                number=-70
+                number=-55
                 [specials]
                     {WEAPON_SPECIAL_PARRY}
                 [/specials]
--- a/units/Delly_start.cfg
+++ b/units/Delly_start.cfg
@@ -406,7 +406,7 @@
                 clone_anim=yes
                 defense_weight=0
                 damage=-20
-                number=-55
+                number=-30
                 [specials]
                     {WEAPON_SPECIAL_PARRY}
                 [/specials]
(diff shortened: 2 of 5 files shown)


Dugy:
One of the things I have learned about game design is that if a tactic is underused, it's better not to be afraid to make it significantly more advantageous. It reduces the probability it will remain underused and if it becomes better than a previous highly advantageous tactic, then it introduces a paradigm shift in the way it's played which is great for replay value.

So my advice is to make even -30%.

Discontinuum:
Well, that makes sense. I was also concerned that such shy attempt won't convince players about worthiness of the tactic. To be honest, I thought you would take a conservative approach to the balance, after spending quite a time with multiplayer folks I got used to the setup where even 1 hp changing of units might lead to fierce rows among balancers

Dugy:
Okay, thanks for the help!

Dugy:
Regarding my approach to balance, LotI is not a PvP setting and doesn't have to follow its practices. Factions don't need to be balanced with each other, maps don't have to be symmetrical. It's more like a chess where you have three queens and five rooks and play against thirty pawns who are aren't there to defeat you but to provide you with some challenge. My goal was to give you a lot of options so that you could test your creativity, think inside the box and make the best of it.

I didn't even expect every unit to be usable, I just threw out a lot of options to see what the players will make out of it. I have created some items intending to enable powerful tactics yet they were seen as mediocre and I have seen mediocre items with a bunch of randomly added modifiers become parts of overpowered tactics. I have seen a lot of emergent tactics that I never expected to be parts of the game. I was impressed by the clever ideas how powerful some units were built. And that was the goal.

Of course, sticking to most effective tactics available harms the replay value. It can't be prevented only by randomised availability of items and restricted access to unit types. And it's a waste that some stuff I have implemented isn't actually used. So overly powerful tactics need to be weakened and unused tactics need to be enhanced, so that a new set of most effective tactics would emerge. And do it again and again. The only limit is that I can't introduce power creep, where the players could steamroll through most scenarios while hoarding potions and other stuff to sacrifice if they come across something actually hard.

In short, it's not the balance that is important. It's the change that is important.

PR #337, opened 2021-08-17
Legacy assignment made as unit_placed instead of post_stats_update

Discontinuum:
The only possible optimization I can think of is to make unit variable "got_legacy" and check that variable first instead of scanning modifications every time. Is it needed so much and will it be significant is debatable

### DIFF
--- a/utils/global_events.cfg
+++ b/utils/global_events.cfg
@@ -2095,6 +2095,12 @@ $soul_eating"
                 name=unit.race
                 not_equals=undead
             [/variable]
+            [and]
+                 [variable]
+                     name=unit.variables.got_legacy
+                     not_equals=yes
+                 [/variable]
+            [/and]
             [then]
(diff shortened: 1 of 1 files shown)


Dugy:
Please do the optimisation you suggest. If you add a check for that variable being set in the unit's variables into the event's filter, it will be far more efficient (that is, `[filter]`, `[not]`, `[filter_wml]`, `[variables]`). That event is called way too often and if there were more events like this one, the game would get noticeably choppy. The scan for legacy advancements might be removed in some later versions when there won't be many saves with legacies added but without the variable.

Discontinuum:
Ah, damn, I didn't read to add as a filter, added in if rn

Discontinuum:
Alright, this way if I got you right

Dugy:
Okay, thanks for the improvement.

PR #338, opened 2021-08-17
Fixing item picking by AI sides in multiplayer

Discontinuum:
So this is my proposal how to fix picking items by AI in mp. We have the "allied_sides" variable. The old DROP macros is implemented backward compatibly, also if "allied_sides" not set the lua code will fall back to the previous mechanism.

### DIFF
--- a/lua/items.lua
+++ b/lua/items.lua
@@ -395,35 +395,64 @@ loti.item.on_the_ground.add = function(item_number, x, y, crafted_sort)
 		halo = loti.item.halo
 	}
 
-	-- Enable "pick item" event when some unit walks onto this hex.
-	-- (see PLACE_ITEM_EVENT for WML version)
-	wesnoth.add_event_handler {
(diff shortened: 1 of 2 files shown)


Dugy:
I think that could work.

However, can this be solved by checking if the side is allied with a human side instead of manually declaring the allied sides?

Discontinuum:
I doubt. First, we have a lot of allies of the human side in campaign whom we don't want to be able to take items. Second, we still have to determine the human side somehow.

I asked about how to tell ai sides from human controlled ones in wesnoth's discord modding channel, they only advised me to add a Custom AI on AI sides... Or to use the controller filter which doesn't work

Dugy:
True. And I think you also said you could not rely on controllers being humans because they could leave in MP.

Okay, is this the definite version I should merge?

Discontinuum:
Yeah

Dugy:
Okay, done.

PR #340, opened 2021-08-18
Disable shuffle in Gladiators MP scenario

Discontinuum:
As it's a coop, no need to have shuffle and problems arising from it


Dugy:
Honestly, I didn't even know there was an option to shuffle sides. But yeah, while it makes sense if multiple players versus multiple players, it's meaningless in coop.

PR #344, opened 2021-08-20
Modularizing GLOBAL_EVENTS and removing hardcoded allied sides

Discontinuum:
For advanced add-on making tweaking global events might prove beneficial, so I think it might be a good idea to split {GLOBAL_EVENTS} a bit into independent blocks, so when an addon maker assembles his global events, he won't have to copypaste all the essential code. A bit changed the order, hope it's not significant.

Also can't find where "recall list items" used, probably it's legacy

And also fixed placed in globals events and utils (gold chests) where players' sides were hardcoded, replaced with $allied_sides as it should be now

### DIFF
--- a/utils/global_events.cfg
+++ b/utils/global_events.cfg
@@ -116,11 +116,7 @@
         first_time_only=no
         [filter]
             [not]
-#ifndef MULTIPLAYER
-                side=1
-#else
-                side=1,2,3,4,5
-#endif
+                side={ALLIES}
             [/not]
         [/filter]
         [check_unit_title]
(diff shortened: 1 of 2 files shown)


Dugy:
Looks okay, but on a scale from 1 to 10, how sure are you that nothing was broken in this process? This has a potential to break really a lot of things.

Discontinuum:
Well, I'm a bit not sure myself ofc. As for the first part, in theory it's almost the same code after macroexpansion. About the second, well, as we started transition from hardcoded player sides, introducing DROPS_EXT, that's a logical step forward.

I tested things a bit, played a test MP scenario, checked force respec and advancing on your and other's turns.

Maybe the first part isn't so necessary (you can copypaste if you need your global events). But the second part -- a bit scary but as I played, nothing seemed broken

Dugy:
I am merging it, let's hope nothing will go unnoticed for more than a few days.

PR #350, opened 2021-10-04
Fallback value for base_seed

Discontinuum:
If somehow base_seed isn't set (old code using LotI, for example), Lua will exit with error "can't apply bitwise operations with nil" and the unit disappears


Dugy:
Thanks for the fix, I didn't think of this.

PR #356, opened 2021-10-30
New API of special notes

Discontinuum:
Wesnoth 1.16 has a new API for special notes, now they are autogenerated from abilities, weapon specials (and ARCANE and SPIRIT are autogenerated too). Autogenerated special notes are shown in a distinct section now, not just in the end of the description. Also it's possible to make a specific special notes for a unit which I did for several ones (duellist wizard, hybrids with tentacles). And I added special notes to abilities they should be drawn from (like temptation for special notes of Succubus, warlord's rule for Duke, etc).

Macros used before now cause a deprecation note to appear. Although this doesn't affect functionality, clogging the log isn't quite good.

And btw I want to address the special note for the demon arcane res: it doesn't seem to be descriptive enough tbh, creatures of magic like spirits and undead are usually highly vulnerable to arcane. Why demons, creatures of magic, are not?


Dugy:
The special notes were a nuisance and it's nice to seem them finally automated.

Thanks for sorting it out.

PR #358, opened 2021-10-31
Fixed zombie rider weapon special AMLAs.

Chergnobarp:
My Zombie Rider has plague (Soulless), greater infect and plague (Monstrosity) all at the same time. The monstrosity plague doesn't work. This seems to fix it.

Note that greater infect is removed when plague (Monstrosity) is added. Not sure if that's desired behavior, but they seem to be incompatible, so that's what I went with.


Dugy:
Thanks.

PR #364, opened 2021-11-03
Replacement of {FORCE_CHANCE_TO_HIT} in order not to reset stats

Discontinuum:
That's a bit clunky, tbh, it turns out I just copypasted the macroexpansion reversing the unit to add object on. Initially I wanted just to place this effect on all bats in "unit placed, side 2, Vampire Bat" event but with it the attack menu shows 100% chances. I don't think we want to show the player so obviously that we tweaked the chances about Vampire Bats, do we?

Things are yet to be fixed with bossed of the Bloodbath, although attacking them is a bit extraordinary event by itself

### DIFF
--- a/scenarios1/09_Escape_from_Oblivion.cfg
+++ b/scenarios1/09_Escape_from_Oblivion.cfg
@@ -202,13 +202,90 @@
         [/message]
     [/event]
     #Let the player get a few zombies to have something to sacrifice.
-    {FORCE_CHANCE_TO_HIT side=1 type="Vampire Bat" 100 ()}
+    #It's not possible to use FORCE_CHANCE_TO_HIT because it resets our equip and amlas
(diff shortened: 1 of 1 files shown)


Dugy:
It would be much less clunky if the object worked only if some variable inside that unit was set (using [filter_self] combined with [filter_wml]) and only set that variable inside the repeated event.

Btw, if you prefer using tabs and break the code style enforced by wesnoth's tools, then you can use the git smudge/clean trick described in the repo's readme. I also prefer tabs and it also has the functional benefit that the add-on server's version of the campaign is significantly smaller on disk.

Discontinuum:
Do you mean setting this effect to Vampire Bats always and then set/unset some variable during an attack of Vampire Bat instead of installing/removing the object?

Dugy:
Yes.

Dugy:
Looks good now. Thanks.

PR #366, opened 2021-11-04
Add a spider zombie to possible variations

Discontinuum:
1.16 added a lot of new zombie types, and I think a spider one fits into some cave scenarios. It has a lot of hp (25/36), but I mentioned it only once in lists so it would be rarely recruited


Dugy:
Good idea.

PR #368, opened 2021-11-05
Retaliation window fix

Discontinuum:
I tested it a bit, now it seems working. The only problem, dunno if it appeared only now or always was here: sometimes if I click fast I feel like some clicks are skipped and a checkbox doesn't change. The next click it works though


Dugy:
I think I recall having earlier the problem you're describing, but it may not be true.

PR #371, opened 2021-11-08
Implemented incubus statues.

Chergnobarp:
I've always thought Lethalia had an unfair advantage over Efraim in statue levels.


Dugy:
I never realised this would not assign a random gender.

PR #372, opened 2021-11-08
Fix variations and Uria's sprites

Discontinuum:


Dugy:
It kinda sucks that the Sylph is missing animations now. This used to be my favourite unit back in wesnoth 1.6, combining good mobility with good offensive powers against undead and the likes.

Discontinuum:
The new sprite is beautiful, but yeah, alas it's not animated at all. 

I've seen the old sprite in the Ooze mini, it retained it even on 1.14, but I dunno if it makes sense to retain them here for us

Dugy:
I'd probably stick with the new one instead of complicating this. Maybe I'll create some animations for this one myself.

Though that one evil Sylph in Chapter II could keep the old sprite because that kinda made that character look like a member of some serious council.

PR #378, opened 2021-11-17
Replacing deprecated FOREACH macro with [foreach] tag

Discontinuum:
With this PR I start replacing deprecated FOREACH (which lacks scoping and overall not that fancy) to [foreach] tag, it's enough common so in order to reduce risks it would be bit-by-bit replacement

### DIFF
--- a/utils/global_events.cfg
+++ b/utils/global_events.cfg
@@ -175,21 +175,25 @@
 #enddef
 
 #define DRAW_ITEM
-    {FOREACH item_list.object p}
-        [if]
-            [variable]
-                name=item_type
-                equals=$item_list.object[$p].number
-            [/variable]
-            [then]
-                [item]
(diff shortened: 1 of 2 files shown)


Dugy:
Maybe it would be good to keep the iterating variable named `i` rather than the random names I was giving to avoid naming conflicts which should no longer be an issue. If you think it's not worth the risk of reintroducing bugs, I can merge it right now.

Discontinuum:
I don't think renaming is worth it, if I miss something during renaming, it will be a bug and potentially a hidden one

Dugy:
Okay. This is quite a risky change in a dynamically typed language.

Discontinuum:
Yes, $this_item is possible, although we should  avoid iterating over it, coz https://github.com/wesnoth/wesnoth/issues/6305 (recently discovered after SXC and Era of Magic, which used this, got broken)

PR #379, opened 2021-11-17
Workaround for "The Library" disappearance bug

Discontinuum:
I reused some empty space in the left-bottom corner to place green and forest resembling the next level and changed to teleporting there. Now Lethalia doesn't vanish


Dugy:
Great, thanks.

PR #384, opened 2021-11-23
{FOREACH} replacement in Ascension and Toxic Sun

Discontinuum:
Tbh don't know if we really need to heal +999


edwardspec:
These `[foreach]` tags are not nested, I assume `$this_unit` can be used in this case?

PR #394, opened 2021-12-18
Enable Soulless to Necrophage through the [modify_unit_type] mechanism

Discontinuum:
post advance now needed only for MP

### DIFF
--- a/extra_advancements.cfg
+++ b/extra_advancements.cfg
@@ -43,7 +43,7 @@
     
     [modify_unit_type]
         type="Soulless"
-        set_advances_to="Monstrosity,Bone Shooter,Revenant"
+        set_advances_to="Monstrosity,Bone Shooter,Revenant,Necrophage"
         set_experience=90
     [/modify_unit_type]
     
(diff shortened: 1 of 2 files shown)


Dugy:
Thanks. I didn't even know that I had allowed Soulless to advance to Necrophage.

Discontinuum:
With Ghoul advancement line I don't understand, to be honest, why Abomination lacks feeding. This unit is viewed as inferior, feeding maybe won't save the situation but at least makes some more sense. Or was it in order to prevent stacking crazy amounts of hp on Abominations?

Dugy:
Yes, exactly. The Ghoul line was no available for long, mainly because the usual _copy & paste body parts_ trick wasn't usable to create a visually distinct level 4 unit, plus the feeding was a balance problem (originally, Preserved Liches had feeding and it was easy to make them nigh immortal with 400 hp, I replaced it by soul eater to deal with the problem). Eventually, I created an animated 3D model that I could render into 2D animated sprites, allowing me to create a proper sprite for a level 4 Ghoul. I didn't give it feeding because it's OP in a long campaign. Necrophate and Ghast still have feeding, but that's not a big deal because they will always advance to Abomination that doesn't have it.

Issue #390, opened 2021-12-03
Strange unit placed event


Discontinuum:
https://github.com/Dugy/Legend_of_the_Invincibles/blob/6d4a732927847aac58f0b684f7416adabba1aaf6/utils/global_events.cfg#L554-L567

Why do we need this? It's triggered only once on the first unit that happen to trigger unit placed. Luckily it seems not affecting that much but it caused a subtle bug in Kill the King on 1.14: when Calcy is modified from Calcy_early to Calcy, it happens that Calcy triggers this event, and UPDATE_STATS prevents him from changing the type. Also interesting that this bug in KtK started to appear after https://github.com/Dugy/Legend_of_the_Invincibles/commit/e88c2149276afc98d7414528835d8f192e4ac2ac It's because before the changes this event was always fired on Event Loader so it didn't affect players' units. And now, with side=$allied_sides, Event Loader created before this variable is set, so the unit placed event fires on the first unit of the player that caused unit placed.

Dugy:
It was meant to make sure every unit would be affected by update_stats before being used. There is a macro to do it for most units appearing in the LotI campaign, but once `unit placed` was added, I added this event to make it happen automatically. However, I forgot that `first_time_only=no` and it seems that some units are negatively affected.

Discontinuum:
first_time_only is impossible to add, as it's an eternal cycle then

Dugy:
The event is supposed to set the `update_started` variable so that the unit would not match the filter next time.

Discontinuum:
Even so, it seems a problem that it prevents {ADVANCE_UNIT id=Calcy Calcy} from succeeding and Calcy stuck as Calcy_early (when I placed other guys ahead of Calcy, this event made them to be bugged the same way)

Dugy:
IIRC, ADVANCE_UNIT is just a `[modify_unit]` tag that changes the unit type. It will trigger `unit placed`, but I don't see why this particular trigger undoes the change. Also, I think it worked fine before 1.16.

Discontinuum:
Yes, it's modify_unit after macroexpansion. No, this problem is with Kill the King on 1.14 with the current LotI

Discontinuum:
Hm, it turns out there is a bug in LotI because of it. If in LotI 2 in the first scenario Lethalia advances, she appears as advanced but then she reappears with backup_amla. And this bug works only if this event is triggered when she advances, so she must not equip any items before it to reproduce
[...]

PR #397, opened 2022-01-05
Don't run initial update_stats on already updated units

Discontinuum:
#390 was still broken for older saves. If the player advanced his AMLA unit without update_started=yes set, it got recalced and the advancement was reverted. Here is a replay with demonstration.

[LotI1-Shadow Empire replay 20220105-062727Bugged.gz](https://github.com/Dugy/Legend_of_the_Invincibles/files/7811704/LotI1-Shadow.Empire.replay.20220105-062727Bugged.gz)

The problem is that unit placed event didn't check "updated=yes", only "update_started=yes". But update_started wasn't set before the first bug-related patch https://github.com/Dugy/Legend_of_the_Invincibles/blame/9929fd0778dfa2c06830c819c16ca6c2272095ee/lua/stats.lua#L45 so in the older saves only "updated=yes" is set but "update_started=yes" is not. But if "updated=yes" is set, it means the unit has been updated at least once so no reason to run initializing update_stats again.

### DIFF
--- a/utils/global_events.cfg
+++ b/utils/global_events.cfg
@@ -536,6 +536,13 @@
                         [/variables]
                     [/filter_wml]
                 [/not]
+                [not]
+                    [filter_wml]
+                        [variables]
+                            updated=yes
+                        [/variables]
+                    [/filter_wml]
+                [/not]
             [/filter]
             {UPDATE_STATS}
         [/event]


Dugy:
Thanks for the fix.

PR #405, opened 2022-01-28
Fix cumulative leadership (considering Wesnoth bug)

Discontinuum:
Wesnoth has a bug (https://github.com/wesnoth/wesnoth/issues/6466 ) in 1.16 that cumulative leadership doesn't work with value= attribute, but works with add=. It can't be fixed in 1.16, only in 1.17, so replacing value= with add= is the option to fix rn for addons.

### DIFF
--- a/units/Flower_of_Evil.cfg
+++ b/units/Flower_of_Evil.cfg
@@ -71,8 +71,8 @@
             {WEAPON_SPECIAL_SLOW}
             [chance_to_hit]
                 id=marksman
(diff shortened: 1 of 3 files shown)


newfrenchy83:
cumulative must be no while what bug not fixed.

Dugy:
Thanks for looking into it.

Discontinuum:
> cumulative must be no while what bug not fixed.
For WC they want to fix it with add= instead of value=

PR #407, opened 2022-02-18
Start uploading beta to 1.17.x server

Discontinuum:
The 1.17 addon server has been opened


edwardspec:
Tools are cached. (because it takes forever to clone Wesnoth repository)
I invalidated the old cache by changing the cache key, it should download new Wesnoth (where tools know about 1.17.x) on the next attempt to upload the beta.

PR #409, opened 2022-04-30
Fix typo "imact" instead of impact

dalong76:
Clearly a bug.


Dugy:
Thanks for spotting that.

PR #411, opened 2022-05-22
Lesser redeem doesn't work on bosses anymore. Fix #408.

### LINKED ISSUE #408: LotI2-The Chase is unwinnable if Krux redeems Deliria
Krux redeemed her at (24, 34). She ran away, but didn't reappear. I double-checked in `:inspect`, she's not hiding, she's truly gone.

Chergnobarp:
This one's untested, since I don't have any save files to work from right now.

### DIFF
--- a/utils/abilities_events.cfg
+++ b/utils/abilities_events.cfg
@@ -5722,7 +5722,7 @@
         [if]
             [variable]
                 name=did_succeed
-                greater_than=0
+                equals=0
             [/variable]
             [not]
                 [variable]
@@ -5730,7 +5730,7 @@
                     contains=$second_unit.id
                 [/variable]
             [/not]
-            [else]
+            [then]
                 [animate_unit]
                     flag=redeem_anim
                     [filter]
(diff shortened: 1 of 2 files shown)


utils/abilities_events.cfg
```diff
@@ -5724,6 +5724,12 @@
                 name=did_succeed
                 greater_than=0
             [/variable]
+            [not]
+                [variable]
+                    name=bosses
+                    contains=$second_unit.id
+                [/variable]
+            [/not]
             [else]
```
Dugy, line 5733: The code around is written quite stupidly, the condition says that if the result of a random number choice is greater than zero, then do nothing, otherwise start the redeem process. With your change, the unit also cannot be on the boss list for nothing to happen, so the result is that the unit is redeemed if it's not a boss or the random number was chosen correctly.

Dugy (commented):
I don't think it can work this way.

utils/abilities_events.cfg
Chergnobarp: Good point. Alright, here's a new commit where I unstupidified the code.

Dugy:
Thannk you.

Issue #414, opened 2022-06-19
Skipping the tutorial doesn't work


Discontinuum:
It says

error wml: loti.item.type[605]: not found in item_list.

I guess it's because the Tutorial doesn't have the item system fully initialized, and item_list isn't loaded

Discontinuum:
In 1.17 Wesnoth Playing tutorial also doesn't work

error scripting/lua: ~add-ons/Legend_of_the_Invincibles_beta/lua/utils.lua:10: bad argument #2 to 'setmetatable' (nil or table expected, got string)
```
stack traceback:
	[C]: in function '.setmetatable'
	~add-ons/Legend_of_the_Invincibles_beta/lua/utils.lua:10: in field 'deepcopy'
	~add-ons/Legend_of_the_Invincibles_beta/lua/utils.lua:8: in field 'deepcopy'
	~add-ons/Legend_of_the_Invincibles_beta/lua/unitdata.lua:104: in field 'item_with_set_effects'
	~add-ons/Legend_of_the_Invincibles_beta/lua/items.lua:582: in field 'describe_item'
	~add-ons/Legend_of_the_Invincibles_beta/lua/items.lua:866: in local 'cmd'
	lua/wml-utils.lua:145: in field 'handle_event_commands'
	lua/wml-flow.lua:5: in function <lua/wml-flow.lua:4>
	[C]: in field 'fire_event'
	lua/wml-tags.lua:152: in local 'cmd'
	lua/wml-utils.lua:145: in field 'handle_event_commands'
	lua/wml/message.lua:469: in local 'cmd'
	lua/wml-utils.lua:145: in field 'handle_event_commands'
	lua/wml-flow.lua:5: in function <lua/wml-flow.lua:4>
```
although in 1.16 (stable) playing tutorial works fine

PR #418, opened 2022-07-19
Fixing skipping tutorial #414

Discontinuum:
The latest fix was only for 1.17 item system bug. The tutorial skip wasn't fixed


Dugy:
I thought I was testing that.

PR #422, opened 2022-08-18
Starter ghost in ch4's An Army Is Born carries over to next scenario.

Chergnobarp:
It was my only wraith, the disappearance was quite noticeable.


Dugy:
Thanks.

PR #423, opened 2022-08-27
Let the user know which redeem attacks can be used in retaliation

Chergnobarp:
I chose arctic blast solely for Efraim to retaliate with. Imagine my surprise when it didn't work. I was two AMLA's in when I realized that it wasn't simply weaker than his fireballs. Alter last advancement saved one, but I now have one wasted AMLA and one wasted redeem path.


Dugy:
Sorry for never mentioning it was attack-only.

Issue #427, opened 2022-11-19
Precious Eagle Cactus Fruit doesn't give heals +40 if its bearer a healer himself


Discontinuum:
Dunno if anything could be done about it, but at least I want to document it here. The superhealing ring won't give its heals +40 if a unit has already some healing ability. So Elvish Shaman with this ring still heals +4, and Druid 8, and so on. If you level into Sorceress, the ring starts working on her

newfrenchy83:
it's because ABILITY_HEALS_OTHER has same id what core healing, when i using diffrerent id and adding via object, it's work but not with same id.

PR #430, opened 2022-11-19
fix Precious Eagle Cactus Fruit doesn't give heals +40 if its bearer a healer himself

newfrenchy83:
fix https://github.com/Dugy/Legend_of_the_Invincibles/issues/427


Dugy:
Thanks.

PR #431, opened 2022-11-19
this fix is more appropriate for ABILITY_HEALS_OTHER VALUE and other

newfrenchy83:
if two ABILITY_HEALS_OTHER VALUE added by oject in same unit thi issue will be the same what in https://github.com/Dugy/Legend_of_the_Invincibles/issues/427.

### DIFF
--- a/utils/abilities.cfg
+++ b/utils/abilities.cfg
@@ -782,7 +782,7 @@ Does not affect poisoned enemies."
 #define ABILITY_HEALS_OTHER VALUE
     [heals]
         value={VALUE}
-        id=healing_other
+        id=healing_{VALUE}
         affect_allies=yes
         cumulative=yes
         name= _ "heals +"+{VALUE}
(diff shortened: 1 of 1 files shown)


Discontinuum:
Indeed, I originally tested with Krux of Chapter 10, and he has heals +16 so the first fix wouldn't helped him

Discontinuum:
Hm, I fear there can be a problem with that. Code for unholy hunger starving checks for "healing" id to determine healers https://github.com/Dugy/Legend_of_the_Invincibles/blob/3ee04819c7269bc2cf8ef4e2d456d85cb4a31af1/utils/abilities_events.cfg#L424
As well as incinerate https://github.com/Dugy/Legend_of_the_Invincibles/blob/3ee04819c7269bc2cf8ef4e2d456d85cb4a31af1/utils/abilities_events.cfg#L2893

So I fear now all the code that checked for healers may stop working for custom LotI heals, and work only for heals+4 and heals+8

Dugy:
I think `ability_type_active=heals`, which will work for any healing that is active at that point. That will include degeneration, though.

Something like `formula="reduce(map(self.abilities.heals, contains_string(self.id, 'heals'), if (a, a, b))"` could make the filter match all units who have healing abilities with `heals` in name, but this is the first time I see the formula thing, so I have no idea if it's correct.

PR #432, opened 2022-11-19
fix 'healing_filtering'

newfrenchy83:
like healing' abilities filtered in abilities_event.cfg, remove existent abilities before add new is only option valid.

### DIFF
--- a/scenarios2/04_Darkness_Closing_In.cfg
+++ b/scenarios2/04_Darkness_Closing_In.cfg
@@ -733,6 +733,14 @@
             [modifications]
                 [object]
                     [effect]
+                        apply_to=remove_ability
+                        [abilities]
+                            [heals]
+                                id=healing
+                            [/heals]
+                        [/abilities]
+                    [/effect]
+                    [effect]
                         apply_to=new_ability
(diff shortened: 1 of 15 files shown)


newfrenchy83:
fix issue reported by @Discontinuum

Dugy:
I am not sure if there isn't a better solution, see my comment at #431 .

newfrenchy83:
i'm not familiar with lua formula and implementation in [filter_adjacent], i cannot repply, and fltering by type include burning abilites and unholybane, the only method for me, and what if you add a new abilites heals, remove the healing already present each time with id only(this is actually only attribute checkedin remove_ability)

Discontinuum:
> I think ability_type_active=heals, which will work for any healing that is active at that point. That will include degeneration, though.

Can we check for the value, whether it's positive or not? I guess degenration abilities have negative value, don't they?

Dugy:
Only by using a lua formula in the filter, I think.

newfrenchy83:
another solution will be to let healing_{VALUE} but to add a heal with id=healing for filtering

PR #434, opened 2022-11-19
Fix great spells resetting stats if not used in chapter 5

Discontinuum:
Instead of writing some extra logic to gracefully remove them, I think we may just update_stats of Efraim and Lethalia at the start of all scenarios? It won't be too much burden, will it?


Dugy:
Thanks for the fix.

PR #436, opened 2022-11-25
Fixing unforgiving ability

Discontinuum:
Unforgiving was broken in several ways. First, it didn't reset the variable after 5 turns. Second, in reality it applied to any attack, not only on those unforgiven unit types. And third, on offense it used [harm_loti] without a harmer, so the unit didn't get XP, soul counts, etc (it was considered that the target just died, not killed by the unforgiving attacker)

Also added a comment for recent healing fix, in order not to forget in the future why it's there


Dugy:
All fine, thanks for the fix.

PR #438, opened 2022-12-29
TBC Goonville: Burke's guards and lawlessness no longer allies

Discontinuum:


Dugy:
Thanks. Good that someone noticed that I put the campaign here.

PR #439, opened 2023-01-06
Default values for some items and Boulderlobber's xp

Discontinuum:
Some stats of items and units depend on the difficulty with QUANTITY macro. But some of them lacked the default number. When used in addons that don't have difficulty levels (so neither EASY, NORMAL or HARD defined) QUANTITY macro doesn't set any value. So items have no effect at all and Boulderlobber has no defined xp threshold for amla


Dugy:
I sometimes forgot to set those, and had no means of checking it. Thanks for the fix.

PR #443, opened 2023-02-04
Made unknown demon infiltrators susceptible to unholybane.

Chergnobarp:
It was a bit of an unpleasant surprise starting ch8 with Eidolon's sword for Efraim.


Dugy:
Thanks.

PR #444, opened 2023-02-04
Lord of the flies weapon special description matches code now.

Chergnobarp:


Dugy:
Oops.

Discontinuum:
Dunno where to ask, we'll ask here. What is the reference in the description of Beelzebub's Hammer "Where is Bruce gone?"? The link between "lord of the flies" and Beelzebub is very obvious, but which Bruce and what does he do with Beelzebub or flies?

Dugy:
Well, _Lord of the Flies_ is also the name of a relatively popular British book (and film adaptations of it). And there also was a song by British band Iron Maiden about it, from an era when they had a different singer than the usual one, _Bruce_ Dickinson.

PR #445, opened 2023-02-19
Close the [unit_type] tag

Bafy78:
the [/unit_type] line at the end of the files was not present for the majority of the unit, which prevented the wmlxgettext file from working, when trying to generate a .pot file for translation of the campaign.

### DIFF
--- a/units/Abomination.cfg
+++ b/units/Abomination.cfg
@@ -327,3 +327,4 @@
             {AMLA_DEFAULT_BONUSES}
         [/advancement]
     ) "Abomination"}
+[/unit_type]
--- a/units/Akula.cfg
+++ b/units/Akula.cfg
@@ -334,3 +334,4 @@
     [/effect]
     {AMLA_DEFAULT_BONUSES}
 [/advancement]) "Akula"}
+[/unit_type]
--- a/units/Ancient_Lich.cfg
+++ b/units/Ancient_Lich.cfg
@@ -365,3 +365,4 @@
         increase=1
     [/effect]
 [/advancement]) "09 Ancient Lich"}
+[/unit_type]
--- a/units/Arch_Necromancer.cfg
+++ b/units/Arch_Necromancer.cfg
@@ -457,3 +457,4 @@
             [/effect]
         [/advancement]
     ) "Arch Necromancer"}
+[/unit_type]
--- a/units/Argan_lich.cfg
+++ b/units/Argan_lich.cfg
@@ -1083,5 +1083,5 @@ Advancements with crossbow add great precision."
             {GREATER_AMLA_DEFAULT_BONUSES}
         [/advancement]
     ) "Argan_lich"}
-
+[/unit_type]
 #undef LIGHTNING
--- a/units/Blackguard.cfg
+++ b/units/Blackguard.cfg
@@ -201,3 +201,4 @@
     {AMLA_DEFAULT_BONUSES}
 [/advancement]
 ) "Blackguard"}
+[/unit_type]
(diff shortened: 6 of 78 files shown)


Dugy:
Does it actually load? Because the ending tags are part of the macro at the end of the files, so adding them would cause duplicate endings of those tags.

Bafy78:
Eh I have no idea how anything works, and you make me realize that not testing something before making a pull request is completely stupid.
I just spent the entire day trying the generate a .pot file to translate the campaign, but the instructions given in the forum were outdated so it was hard :/ (I will update them)
All I know is that the file wouldn't generate without my adding these lines, so I thought that it would be better to make a pull request, for the future people who would want to also translate the campaign.
I just tried loading a save of the campaign, and there's no directly apparent problem to me (in this save I have units which files I edited).
I am not accustomed to using github, sorry for this.

Dugy:
You probably didn't try to run the folder that you had edited. The campaign with your changes does not run for me.

Once you sort that folder problem out and it starts failing with an error, you can fix it by adding that line to every unit file that contains the `GENERIC_AMLA` macro and removing that tag from the declaration of the `GENERIC_AMLA` macro from the `utils/amla.cfg` file.

Discontinuum:
It prevents the new python wmlgettext but old perl wmlgettext should work.

The closing tag isn't put because that's how advancing macros are done, to change it we need to change macros a bit. Is it worth it?

Bafy78:
oh wait, I am stupid: I edited the amla.cfg file too, but I forgot to add it.

Dugy:
You forgot to add many other files as well, I think. The usual trick to prevent forgetting that is to always do `git commit -a`...

Bafy78:
> It prevents the new python wmlgettext but old perl wmlgettext should work.

Well I didn't succeed in running the old wmlxgettext, so when I saw there was a new one, I used it.

Dugy:
> The closing tag isn't put because that's how advancing macros are done, to change it we need to change macros a bit. Is it worth it?

I think yes. If the Python tool is what is used now, it's better if it works with that.

Bafy78:
> You forgot to add many other files as well, I think. The usual trick to prevent forgetting that is to always do `git commit -a`...
No, I didn't forget any other files: I still have the log open of which files were problematics, there are only units and amla.cfg 
And actually I have no idea how to use git, but directly uploaded the files to github haha, I guess I will need to learn how it works. What should I do now to add the amla.cfg to the pull request? (if it's possible)

Dugy:
Not all of the units were added, only those up to `Lethalia_lich.cfg`. So you'll have to add them as well.

Because this is not my pull request, I don't see the same buttons as you do, but you should got to the https://github.com/Bafy78/Legend_of_the_Invincibles branch and upload the rest of the files there.

Dugy:
That is a step in the right directions, but you haven't added the changes to all of the unit types.

Bafy78:
Sorry, I really thought I uploaded them all. But I just saw there is a 100 files restriction when uploading. As I uploaded every file (I didn't remember which ones I edited), I reached this limit without noticing.

Dugy:
Yeah, I thought it was something like that.

Anyway, I merged it into the master branch, so this problem should not be happening to anymore now.

Bafy78:
Thanks! I will work on the translation now.

Dugy:
What translation, by the way?

Bafy78:
In french! I saw that someone said that they were working on it, in 2019, but no further messages, and he doesn't answer. So I'll start from the beginning, unless you know someone is working on it rn.

Dugy:
I don't know of any French translation being done. Thanks for looking into it, it might help me restore my long unused French language skills.

PR #446, opened 2023-02-22
Advancing- unit types were not created

Discontinuum:
The latest change to amla.cfg closed the tags in the wrong order, so during the macroexpansion it was like
[unit_type]
id=Type
[unit_type]
id=AdvancingType
[/unit_type]
[/unit_type]
so AdvancingType didn't get created and advancement didn't work

изображение

### DIFF
--- a/utils/amla.cfg
+++ b/utils/amla.cfg
@@ -64,6 +64,7 @@
         require_amla=""
     [/advancement]
 #wmllint: unbalanced-off
+[/unit_type]
 [unit_type]
     id=Advancing{TYPE}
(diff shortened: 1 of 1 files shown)


Discontinuum:
although I wonder if now getting .pot doesn't work again, as now amla.cfg has unbalanced tags

Dugy:
Thanks for fixing it.

> although I wonder if now getting .pot doesn't work again, as now amla.cfg has unbalanced tags

There is the comment about unbalanced tags, I hope gettext supports that.

PR #447, opened 2023-03-17
Loyal icon superimposed and prestart messages fixes

Toranks:
These icon overlays are a problem caused either by the loyal code without an empty overlay, or by using IS_HERO and TRAIT_LOYAL together, existing TRAIT_LOYAL_HERO, better and more specific.  And also because of the lua code that adds the loyal icon to all loyal units without checking if they are also heroes. The easiest way to spot it is to check if it has the hero ellipse effect.
Once the first two scenarios are resolved for Delenia and Mario , if the changes seem good to you, I will do it with other characters with the same problem as I progress in the campaign, since I like that the icons on characters that will accompany me in many chapters look beautiful.

imagen


Dugy:
Feel free to continue, I will merge it then. I never paid much attention to these icons and I didn't notice it didn't always work as it should have.

Toranks:
Extra check to avoid give loyal icon to a leader, who already has a crown.

Toranks:
I'll continue here the same cases as here that I find myself throughout the campaign:
https://github.com/Dugy/Legend_of_the_Invincibles/pull/456

Toranks:
@Dugy see the last commit. The shop only allows you to buy 1 item and you have to move the unit to one adjacent square to buy again. As the solution was simple, enclosing the message in a while until "Nothing" is clicked, I have done it for all shops. What do you think, do I keep those changes, or is the original intention for the shops to be for 1 use only?

Dugy:
Yeah, keep that. I thought about giving it a nicer GUI, but I didn't get myself to do it.

Toranks:
@Dugy on the last commit I have fixed a problem that makes Vlad attack in the opposite direction of the victim and it looks bad. I guess that you are okay with this, it's a little fix.

Dugy:
I don't have the rights to view anything in that Discord channel.

Toranks:
See this;
https://github.com/wesnoth/wesnoth/issues/6258
https://github.com/knyghtmare/Struggle_For_Freedom/pull/7

Dugy:
Yeah, I am fine with that addition.

Toranks:
Last commit:
imagen
Before the text was so long that it was not visible on the screen

Toranks:
I have finished the entire part 1, it is ready for review, all the changes have been checked throughout these months.

Dugy:
Looks okay.

You've done two unrelated things in a large number of commits in a single pull request. I am not asking you to split it up because it would be really annoying work, but next time please make two pull requests for things like this.

I've merged it all now.

PR #448, opened 2023-03-17
pot file generated and necessary wmlxgettext fixes

Toranks:
I did this to see how many strings the addon had. (10,000!). In case it helps, to prevent wmlxgettext from giving errors when regenerating the pot.


Dugy:
Generated files like `.pot` should not be part of the repository because they'll become out of date when some sources change.

The other changes are okay.

PR #453, opened 2023-03-20
[set_specials] mode= mandatory

Toranks:
This fixes this spam error on screen (only visible with debug mode):

imagen

Also, the default mode is **replace**, and must be **append**, so this can fix potentially errors with different specials on the same weapon.


Dugy:
This will probably keep annoying for quite a while, since it seems that these warnings manifest only when the part of the code is executed. If this is the case, it may be necessary to figure out a better solution than a dozen merge requests.

Toranks:
What do you mean? It's solved now, it was just an oversight when writing the [set_specials] tag I guess. I have done a text search of all the set_specials in the campaign and all are fine except these two.

Dugy:
Okay then, I thought you just found those two and many more could be lying there, undiscovered.

Toranks:
Oh no, I already searched all

PR #455, opened 2023-03-21
Save Corruptia

Discontinuum:
Corruptia may be killed in the last battle. But her book is mentioned in Chapter 4, in the library. Instead of making it conditional or silently glossing over and supposing she survived somehow, maybe just save her the same way Delly is saved? Efraim and Lethalia have an incentive to keep her live: she's the only surviving leader of Arach, so she may help restoring the order after Lilith is defeated

### DIFF
--- a/scenarios2/10_Lilith.cfg
+++ b/scenarios2/10_Lilith.cfg
@@ -2052,6 +2052,31 @@ As the time comes for two suns to shine brightly in the sky."
             message= _ "Not until I allow it! This should prevent her from dying. We will worry about her later."
         [/message]
     [/event]
+    [event]
+        name=last breath
+        [filter]
+            id=Corruptia
+        [/filter]
+        [message]
+            speaker=unit
+            message= _ "No! Save me! Curse you, Lilith!"
+        [/message]
+        [message]
+            speaker=Efraim
+            message= _ "Among the leaders of Arach she was the only one who turned away from Lilith. She might be helpful to bring local people back to normal life."
+        [/message]
+        [message]
+            speaker=Lethalia
+            message= _ "Anghra-kha'sambghra, sei'ko osstap nalatho, ghelt dob'the ekthanathul."
+        [/message]
+        {MODIFY_UNIT id=Corruptia status.petrified yes}
+        {MODIFY_UNIT id=Corruptia hitpoints 40}
+        {MOVE_UNIT id=Corruptia 42 5}
+        [message]
+            speaker=Lethalia
+            message= _ "I agree. If she dies here, we'll have to restore the order ourselves. Without any help from old leaders it will take a lot of time and effort. This should prevent her from dying. We will worry about her later."
+        [/message]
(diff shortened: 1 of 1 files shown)


Dugy:
She almost always died when I play tested that scenario and I always assumed that she was dead in the next chapters. She could have written the book before or after the battle in Arach, and it had to be before in case she died.

Also, I never intended her to be a good person. Just more pragmatic than the other villains (though your dialogue reflects that). Do you really think adding a mention the protagonists saved her would be appropriate?

scenarios2/10_Lilith.cfg
```diff
@@ -2052,6 +2052,31 @@ As the time comes for two suns to shine brightly in the sky."
...
+        [filter]
+            id=Corruptia
+        [/filter]
+        [message]
+            speaker=unit
+            message= _ "No! Save me! Curse you, Lilith!"
+        [/message]
+        [message]
+            speaker=Efraim
+            message= _ "Among the leaders of Arach she was the only one who turned away from Lilith. She might be helpful to bring local people back to normal life."
```
Dugy, line 2066: _From all of Arach's council members, she was the only one who turned away from Lilith. She might turn out to be helpful at bringing normal life to the city._

scenarios2/10_Lilith.cfg
```diff
@@ -2052,6 +2052,31 @@ As the time comes for two suns to shine brightly in the sky."
...
+        [message]
+            speaker=Lethalia
+            message= _ "Anghra-kha'sambghra, sei'ko osstap nalatho, ghelt dob'the ekthanathul."
+        [/message]
+        {MODIFY_UNIT id=Corruptia status.petrified yes}
+        {MODIFY_UNIT id=Corruptia hitpoints 40}
+        {MOVE_UNIT id=Corruptia 42 5}
+        [message]
+            speaker=Lethalia
+            message= _ "I agree. If she dies here, we'll have to restore the order ourselves. Without any help from old leaders it will take a lot of time and effort. This should prevent her from dying. We will worry about her later."
```
Dugy, line 2077: _It could take a lot of time and effort without the help of the old leadership._

Dugy (changes requested):
The English seems a bit off.

Discontinuum:
> and it had to be before in case she died.

I admit I didn't think of it, which of course might've been the case.

> Do you really think adding a mention the protagonists saved her would be appropriate?

I'll try to rephrase it (alongside with your edits). I think a mention that they saved, or rather spared her, may be worth it because it fits into the general narrative of Part 1 that Efraim and Lethalia did many questionable things with the intent of good. Corruptia isn't good and just more pragmatic, but Lethalia is very pragmatic herself in her approaches of doing good. This may be one of the questionable things heroes did (and quite innocent compared to what they've done in Part 4 and 5)

Dugy:
I have edited it myself a bit and pushed into master myself.

PR #456, opened 2023-03-24
S08 moved messages from prestart to start

Toranks:
Two messages have been sneaked into the prestart event, which has this undesired effect. Moved to start, immediately after.

imagen


Dugy:
This wasn't causing problems when I was writing it back on Wesnoth 1.6.

PR #461, opened 2023-04-02
Make Efraim hidden on P1-C2-S3

Toranks:
As you can see in the image, Efraim is present on this scenario at the bottom, visible if you zoom out or after the victory event. And it's not supposed to be there, not even in secret. With this PR, he will be hidden all the time, from prestart to victory event (and then go straight to recall).

imagen


Dugy:
Thanks. I didn't know such tricks at the time I was writing this scenario.

PR #463, opened 2023-04-04
Premature message on "Umbra" scenario

Toranks:
This message:
`"Damn those two liches! They will make this battle turn against us!"`
Can be triggered by the dwarves, before the liches has a chance to enter the caverns (with luck, even before the arrival of Umbra). It makes more sense if it's a unit from side 1 that triggers the message and turn the tables.


Dugy:
Yeah, that makes sense.

PR #464, opened 2023-04-05
Unslowable and undrainable instead of event-based

Toranks:
It avoids the need to have several extra events by using a status recognized by the engine and trusted, but not documented on the wiki.

Still requires test these specific objects, so draft until someone else can confirm at least one of these items. But the solution is simple and generic, I doubt it won't work for every item.

https://github.com/wesnoth/wesnoth/pull/5788

https://wiki.wesnoth.org/SingleUnitWML#Unit_State


Dugy:
Yeah, this wasn't possible before, but the more recent additions have enabled to get rid of the overcomplicated event-based workaround.

Toranks:
The Unraveller tested. Lethalia does not receive slow and Efraim does, from wolves with slow (created with debug mode)
imagen

PR #465, opened 2023-04-08
Changed special= for special_id or special_type

Toranks:
special_id refers to id, and special_type to the tag [plague] (the only case on LOTI).
This message appears on every attack by some characters in debug mode:
imagen

Don't worry about compatibility, all three keys have existed since they were created, just special= will be deprecated


Dugy:
Thanks for looking into it. Are you sure these are all the occurrences of the deprecated code?

Toranks:
> Thanks for looking into it. Are you sure these are all the occurrences of the deprecated code?

Yes, I did a string search for all files

Dugy:
All right. Thank you for fixing it.

PR #467, opened 2023-04-14
Clear shroud on Ch3 S01 final conversation

Toranks:
Without this, the conversation will take place by talking to invisible units if there is no unit near the human city.
Also changed trait_loyal + is_hero with trait_loyal_hero on this scenario

imagen


Dugy:
Thanks.

PR #468, opened 2023-04-15
Fix moveto onto Ziggurat only working the first time

matsjoyce:
Because of my lack of situational awareness (I didn't notice the shops :smile:), I ran into a bug where you only get the option to enter the ziggurat once. This should fix it (I haven't tested it, but referencing the rest of the file and my rusty WML knowledge give me reasonable confidence this works).


Dugy (approved):

Dugy:
This is not the first and probably also not the last case when I have forgotten about the `first_time_only=no` attribute.

PR #469, opened 2023-04-16
unit.side instead of side_number on attack related events

Toranks:
This is more reliable on attacks events. Fine on side events, but when active attack is caused by some kind of event or ability, `side_number `is the active side, not the attacker side.
I don't know if there is any event or ability like this in all of LOTI that can interact in this way, since I detected the error with my own addon (Advanced Wesnoth Wars ambush feature), but it doesn't hurt to trust the variables used in these events.

You can compare the before and after here;
Using $side_number

https://user-images.githubusercontent.com/40789364/232260685-38e5a7b7-3d60-4b6d-9820-473fc5625876.mp4

Using $unit.side (don't worry about the lua warning, this video is recorded in debug mode and always appear with [harm_unit_loti])

https://user-images.githubusercontent.com/40789364/232260691-e54b8612-1231-42ee-ba2e-4c9b0d0e1401.mp4

All attack related events have been checked so that the correct variable side is used, **_dazzled_** uses `second_unit.side` for **`defender hits`** event.

### DIFF
--- a/utils/abilities_events.cfg
+++ b/utils/abilities_events.cfg
@@ -584,7 +584,7 @@
                 # We do not hurt our own units.
                 [filter_side]
                     [enemy_of]
-                        side=$side_number
+                        side=$unit.side
                     [/enemy_of]
                 [/filter_side]
                 # The defender was already damaged.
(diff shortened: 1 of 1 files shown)


Dugy:
I didn't realise this problem could happen. Thanks for figuring it out.

PR #470, opened 2023-04-27
Fixed hotkey for unit info

Toranks:
Have you noticed that when you press shift+c and select a unit, you get an error message? So, the hotkey to "Unit info" doesn't work. The error was simple, the filter was badly placed.


Dugy:
Thanks for the fix.

PR #471, opened 2023-04-28
Fixed incinerate on defense

Toranks:
In rare cases, when the unit is already dead and the incinerate event try to do unstore_unit here, you get an error on the screen and various errors in the log:

```
20230428 15:43:48 error engine: failed to auto-store $unit at (39,26)
20230428 15:43:48 error engine: failed to auto-store $unit at (39,26)
20230428 15:43:48 error wml: [unstore_unit]: variable 'unit' doesn't contain unit data

20230428 15:43:48 error engine: failed to auto-store $unit at (39,26)
20230428 15:43:48 error engine: failed to auto-store $unit at (39,26)
20230428 15:43:48 error engine: failed to auto-store $unit at (39,26)
20230428 15:43:48 error engine: failed to auto-store $unit at (39,26)
```

To avoid this, simply use the same method as in the attacker_hits incinerate event, with the conditional
```
         [if]
             [have_unit]
                 find_in=unit
             [/have_unit]
```

[have_unit] will filter out units that are still alive after the attack. If the unit is died, the filter will do negative and it will stop.

### DIFF
--- a/utils/abilities_events.cfg
+++ b/utils/abilities_events.cfg
@@ -2839,45 +2839,52 @@
         [/filter_second_attack]
         {VARIABLE unit.status.incinerated yes}
         {VARIABLE unit.variables.incinerator $second_unit.id}
-        [unstore_unit]
-            variable=unit
-            find_vacant=no
-        [/unstore_unit]
-
-        {VARIABLE is_red no}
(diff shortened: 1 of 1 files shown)


Dugy:
Thanks for fixing it.

PR #474, opened 2023-05-09
Necromancy always takes precedence over infected

Toranks:
In the case that an Arch Necromancer infects another Arch Necromancer, and the Arch Necromancer dies from the poison on turn_refresh, it becomes walking corpse instead of a lich. This does not happen if he dies from other causes while infected, so it should be considered a bug. I assume necromancy should take precedence.
I also delete two unnecessary fire_event when unit becomes walking corpse. These events are executed as if the killed unit was the walking corpse, when it should be the necromancer. However, these events are not needed in this situation.

### DIFF
--- a/utils/abilities_events.cfg
+++ b/utils/abilities_events.cfg
@@ -2702,6 +2702,7 @@
             [/unstore_unit]
         {NEXT i}
         {CLEAR_VARIABLE unzombie_store}
+# To cover the case where the infected unit is a necromancer
         [store_unit]
             [filter]
                 status=infected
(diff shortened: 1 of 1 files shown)


Dugy:
Is the last event necessary? If the infected status does not trigger, the necromancy ability will, and to its job, no?

Toranks:
> Is the last event necessary? If the infected status does not trigger, the necromancy ability will, and to its job, no?

If I don't activate it, the unit won't die, since poison doesn't kill. But it is supposed that infection CAN kill. If it has 1 HP after being poisoned, it dies according to the code of this turn_refresh event.

Dugy:
Aha, I see now.

PR #479, opened 2023-05-22
Vampire general advances to Vampire overlord

Toranks:
As it is now, it advances itself and produces a loop.
And strangely, Wesnoth crashes when you look the unit with GUI Debug Tools.


Dugy:
Thanks for the fix.

PR #480, opened 2023-05-23
Delly is a female Preserved lich

Toranks:
Self explanatory

### DIFF
--- a/units/Delly_lich.cfg
+++ b/units/Delly_lich.cfg
@@ -53,7 +53,7 @@
 #enddef
 [unit_type]
     id=Delly_lich
-    name= _ "Preserved Lich"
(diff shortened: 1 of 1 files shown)


Dugy:
I wonder if there is a language where _lich_ has a female form, but well, it's more correct this way.

Toranks:
Lich>Liche on both forms, but "Preserved" has two forms on male and female. Preservado/Preservada

Dugy:
This _liche_ can be grammatically both male and female depending on the context?

Toranks:
> This _liche_ can be grammatically both male and female depending on the context?

Yes

Dugy:
Fair enough.

PR #481, opened 2023-05-27
Secondary unit on kill events

Toranks:
Two fixes:
[filter_second] isn't a valid tag on [kill] (no effect deleting them).
Added [secondary_unit] to [kill]. Without this, the unit killed isn't reflected in statistics.
This does not give experience to the assassin (For that is used [harm_unit] with kill=yes), so the mechanic remains the same.


Dugy:
Good observation.

PR #482, opened 2023-05-27
mass_produced variable on ghost+walking corpse

Toranks:
When a walking corpse is spawned on a plague victim in these scenarios, a ghost is additionally created, but it does not receive the mass_produced variable, resulting in it ending up on the recall list instead of being summoned directly.

imagen


Dugy:
Good observation.

PR #483, opened 2023-05-28
prevamp_store restore HP and status

Toranks:
Without this, units are recalled with the status and HP they had when saved to the variable.

infected and petrified are there just in case, but probably not needed


Toranks:
@Dugy there is a typo `$this_item.max_hitpoints` must be `$prevamp_store[$i].max_hitpoints`, can you change this directly instead of open another PR?

Dugy:
Okay, done. I thought it was some fancy new thing added to WML in 1.16.

PR #485, opened 2023-06-06
Recall new advancement Fire Wraith

Toranks:
Also, the chapter 5 scenario 1 FAQ says that fire guardians don't level up, which is not true, so I'm assuming there was an update time ago where you forgot about these changes.


Discontinuum:
In your commit you recall Fire Guardian the second time, not Fire Wraith

Discontinuum:
Also, why not remove "they will never advance to anything" phrase from the walk-through, if you noticed it?

PR #486, opened 2023-06-06
Remove the lose condition (Death of Efraim or Lethalia) in the Last Crusade from some point

Discontinuum:
I've noticed recently several complaints that people not getting how to break the cycle in the Last Crusade and proceed on. Someone proposed to remove "Death of Efraim or Lethalia" from lose conditions after some time to make it more obvious. I removed it after the 9th cycle, but maybe you think the other time will be better (if you agree on this change)

### DIFF
--- a/utils/chapter9_utils.cfg
+++ b/utils/chapter9_utils.cfg
@@ -1541,6 +1541,14 @@
         [objective]
             description=_ "Death of Efraim or Lethalia"
             condition=lose
+            [show_if]
+                [not]
+                    [variable]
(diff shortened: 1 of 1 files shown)


Dugy:
Yes, the silent disappearance of the lose condition could be a useful hint for those who notice.

Discontinuum:
Your comment sounds approving but you haven't merged. Is anything wrong?

Dugy:
Sorry, I have forgotten.

PR #488, opened 2023-06-11
Remove incineration on thankful_enemy

Toranks:
I noticed that the ally appears already incinerated, the effect she had when I killed her, so I added the same deletion that does the event where the incineration heals.


Dugy:
Thanks.

PR #491, opened 2023-06-12
Option to enlarge map on chapter 5

Toranks:
What do you think about this? I did it for myself, for comfort. It will only look like this in 1920*1080, but it's optional.

https://github.com/Dugy/Legend_of_the_Invincibles/assets/40789364/110a4cc0-5055-4bf1-a30b-a3736e52cc9c


Dugy:
Thanks.

PR #493, opened 2023-06-14
Update [print] duration

Toranks:
> (Before 1.15.4) This is measured in the number of 'frames', and the default is 50. A frame in Wesnoth is usually displayed for around 30ms.
> ([Version 1.15.4 and later only](https://wiki.wesnoth.org/DevFeature)) This is measured in milliseconds.

So 100 ms is clearly insufficient and there is barely time to see anything. Updated to the new default, 5 seconds (5000 ms)
If you prefer another duration, tell me


Dugy:
Good that you noticed. They could have the decency to create a new key named `duration_ms` to avoid causing silent bugs like this.

PR #495, opened 2023-06-18
Fixed tripled enemy id

Toranks:
All these scenarios have the same serious error. As all the enemy bosses have the same id, only the last one appears.


Dugy:
Strange, this error must have gone unnoticed for 10 years.

PR #497, opened 2023-06-19
Soulgate: spawn random units on passable hex

Toranks:
Simple fix. I realized this when I summoned all the Valhalla units.


Dugy:
I think there might be a risk the units would be spawned behind some wall that should be hard to get around, but I guess there are some tricks for doing that already.

PR #500, opened 2023-06-21
[harm_unit] type= > damage_type= fix

Toranks:
There are numerous places where `type=` is used instead of `damage_type=` in `[harm_unit]`, so it has no effect and always does the same damage.


Dugy:
Wow, how do you even notice those?

Toranks:
Because the tentacles of the grand marshall and royal guard did the same damage to me on all units, and I did a search in all the code to see if it happened in more places.

Dugy:
Still, it was an interesting observation.

PR #501, opened 2023-06-21
Add tentacle special note to Grand Marshall too

Toranks:
Another little cosmetic fix. The Grand Marshall has no visible indication of its tentacle power, so it adds a special note just like the royal guard.

### DIFF
--- a/units/hybrids_of_steel.cfg
+++ b/units/hybrids_of_steel.cfg
@@ -524,6 +524,9 @@
     cost=66
     usage=fighter
     description= _ "These are the leaders of the Steel Future society."
+    [special_note]
+        note= {SPECIAL_NOTES_TENTACLES}
+    [/special_note]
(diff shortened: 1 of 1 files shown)


Toranks:
All human-mechanical units have halo with tentacles

Toranks:
This one also has a halo with tentacles, but it also has the power to do 10 damage with each melee attack, exclusive to the grand marshall and the royal guard

Toranks:
Here is the event:
https://github.com/Dugy/Legend_of_the_Invincibles/blob/a9084294c78fa7ac2e10be1d8d94d8e788965119/utils/chapter5_utils.cfg#L1141

Dugy:
I think I misunderstood the comment. The general has that tentacle animation, just like any other, but the information is just missing from the help text?

Toranks:
I think it has a special defense animation, but it's barely noticeable. Anyway that's not the point. Look at the event that I linked to you before, that is what the special note in grand marshall and royal guard of the steel hybrids refers to

PR #502, opened 2023-06-22
Prisoner Nightstalk fix

Toranks:
One of the prisoners has a modification similar to the Nightstalk object on a weapon he does not own. Deducing from the code and the description, it is obvious that it refers to "shadow wave", which already has a normal backstab, like the one the code intend to replace.

### DIFF
--- a/scenarios5/25_Prison.cfg
+++ b/scenarios5/25_Prison.cfg
@@ -174,7 +174,7 @@
                     number=50
                     [effect]
                         apply_to=attack
-                        name=dagger
+                        name=shadow wave
                         remove_specials=backstab
                         [set_specials]
                             mode=append


Toranks:
Don't work. The item is automatically replaced by the "Nightstalk" dagger, which has no effect on any Shadowwalker weapon.

imagen

Toranks:
Another detail is that the description:
`"This dagger creates shadows by itself, allows its user to stalk in the shadows, darkens and backstabs very effectively."`
It is unique to this weapon but never appears in the item window once the unit is handed over to the player, due to the code replacing it with the original Nightstalk dagger.

Dugy:
The object was copypasted into the unit and its code became very obsolete since then.

It was supposed to apply to its dagger attack, replace backstab with greater backstab and add the darkens and nightstalk abilities, so that the wearer would be invisible in darkness but the darkness itself was created around the unit (that is meant by the _shadows it creates by itself_). So it does what was it supposed to do.

However, the item was changed quite noticeably in the items list since the time it was copied into that scenario's code and there are some inconsistencies which can be considered a bug.

Toranks:
Yeah, but you understand that shadowalker uses swords and crossbow, not daggers, right?

Dugy:
He used to use daggers as well. Very long ago.

PR #503, opened 2023-06-22
Inventory

white-haired-uncle:
Added some sorting of units/items to make dealing with inventory a little easier.

My first time using git, not sure what I'm doing here yet.


Dugy:
Yeah, good idea, that could be helpful to everyone.

PR #504, opened 2023-06-22
Make Configure Autorecall menu more user-friendly

white-haired-uncle:
- remove units that are already on the autorecall list from the list of units to add
 - sort autorecall list, units by name (with blank names at end), then by unit type
 - show "not enough gold to add to recall" message
 - keep Configure Autorecall open until you're done
 - add numbers/oblique to units on recall list to differentiate those that will be recalled

I was going to add a new option to move existing units up/down in the autorecall list, which I think would be a big help.  But then I realized that would be enough work that I really should just go all the way and replace the existing menu with:

Initially, show Back/Add Slot using Gold/Select Unit When user selects unit, new options appear:
 - Move unit up/down
 - Add unit(s) above
 - Add unit(s) below
 - Remove unit

Consistent, easy to use, and way more WML than I care to take on.

### DIFF
--- a/utils/chapter9_utils.cfg
+++ b/utils/chapter9_utils.cfg
@@ -1541,14 +1541,6 @@
         [objective]
             description=_ "Death of Efraim or Lethalia"
             condition=lose
-            [show_if]
-                [not]
-                    [variable]
-                        name=crusade_difficulty
-                        greater_than_equal_to=10
-                    [/variable]
-                [/not]
-            [/show_if]
         [/objective]
         [objective]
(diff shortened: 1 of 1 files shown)


Dugy:
The diff is totally messed up, I suspect you have committed some random changes you didn't want to commit.

I can see that some lua code was added, but it's hard to tell what is an addition and what is just messed up diff.

white-haired-uncle:
I figured there were probably too many changes in there.  I kept wanting to do "just one more thing".  I'll break up the changes.

PR #505, opened 2023-06-22
Sort add units, not enough gold message

white-haired-uncle:
- Sort the list of units when adding units.  First by name, with the units without names last (assume most players are less interested in unnamed undead units), then by unit type.
- When you don't have enough gold to add to the list, present a message telling you how much you need instead of just not showing the option to add.
- Replaced one {FOREACH} with [foreach] since I was there

### DIFF
--- a/utils/chapter9_utils.cfg
+++ b/utils/chapter9_utils.cfg
@@ -2287,7 +2287,36 @@
                                         kill=no
                                     [/store_unit]
                                     #Add them to the list
-                                    {FOREACH unit_list i}
+
+# Begin sort add unit list
+[lua]   
+    code=<<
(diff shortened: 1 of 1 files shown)


Dugy:
Yeah, now I can see what you actually changed.

Next time, please indent the WML code as the code around it (lua itself is okay).

white-haired-uncle:
Is using wmlindent okay?  

I suck at keeping things lined up.  Tried to go back and use emacs again, but it's been way, way too long.

thanks

Dugy:
Yeah, wmlindent should be okay. I am not sure how it handles inline lua, but there's nothing better.

PR #506, opened 2023-06-23
Keep Configure Autorecall menu open

white-haired-uncle:
If you want to add a unit and remove a unit, for example, you shouldn't need to open Configure Autorecall multiple times.

Probably should rename top level Back button to Exit

### DIFF
--- a/utils/chapter9_utils.cfg
+++ b/utils/chapter9_utils.cfg
@@ -2151,146 +2151,153 @@
             id=5confautorecall
             description=_ "Configure Autorecall"
             [command]
-                [if]
+                {VARIABLE conf_autorecall_done no}
+                [while]
                     [variable]
-                        name=max_autorecall
-                        greater_than_equal_to=8
+                        name=conf_autorecall_done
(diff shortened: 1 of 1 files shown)


white-haired-uncle:
I don't know why it looks like so much has changed in that one file.  It only happens after running wmlindent.  Without it:
```
spartan$ diff chapter9_utils.cfg.b4indent chapter9_utils.cfg.master 
2154,2161d2153
<                 {VARIABLE conf_autorecall_done no}
<                 [while]
<                     [variable]
<                         name=conf_autorecall_done
<                         equals=no
<                     [/variable]
<                     [do]
< 
2477d2468
<                             {VARIABLE conf_autorecall_done yes}
2482,2484d2472
<                     [/do]   # configure autorecall
<                  [/while]
<                  {CLEAR_VARIABLE conf_autorecall_done}

```

Dugy:
All right, I have done that change manually and committed it myself. The problem was that something in the file was wrongly indented, causing wmlindent to make additional changes.

PR #507, opened 2023-06-23
Remove units from add list already on autorecall list

white-haired-uncle:
Currently, "Add units" includes units that are already on the autorecall list.

There's probably a more elegant way in lua to check if an object is already a member than iterating through the lists, but I couldn't find it.


Dugy:
> There's probably a more elegant way in lua to check if an object is already a member than iterating through the lists, but I couldn't find it.

I don't know, but I would not be surprised if there wasn't, lua doesn't have many library functions, its main selling point is its overall small size.

PR #509, opened 2023-06-23
Autorecall ordered oblique

white-haired-uncle:
Present autorecall list as ordered list, with units that won't be recalled flagged with oblique font.

I'm a little embarrassed by the "my_message" stuff.  Worked well enough to minimize changes to existing code while testing.  When I tried to move the if i < max_autorecall conditional inside the value block it didn't work any more, and I just don't care enough about style to beat my head against wml to make it pretty.

Trying something new (to me) with branches here, hopefully it makes sense, otherwise I can redo.

### DIFF
--- a/utils/chapter9_utils.cfg
+++ b/utils/chapter9_utils.cfg
@@ -2294,7 +2294,30 @@
                                                 kill=no
                                             [/store_unit]
                                             #Add them to the list
+                                            # Begin remove units on autorecall from list of units to add
+                                            [lua]
+                                                code=<<
+        local ul = wml.array_access.get "unit_list"
+        local ar = wml.array_access.get "autorecall"
+        local newul = {}
 
+        for _,a in pairs(ul) do
+            local found_match = 0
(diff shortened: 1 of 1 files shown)


Dugy:
This looks like a candidate for being reconstructed as a lua-based UI, somilarly to the inventory.

Toranks:
If I see that the map does narration spoilers, I can convert them to story messages with a flat black background.
Anyway, most of the ones I've seen so far go right after the story screen (which has its own narration), so it doesn't matter that the narrator speaks during the beginning of the scenario.

white-haired-uncle:
I've had a couple ideas about how I would design the whole autorecall menu it I were to write it from the ground up with the benefit of a good bit of hindsight from experience as a player.  

But, it's not a big part of the game like the inventory, and IMO it's just about good enough (besides breaking the menu out and sharing it between ch5&9, this is probably the last of my tweaks except maybe adding 'geared' to the unit lists).

PR #510, opened 2023-06-23
Part II fixes

Toranks:
- Messages prestart to start (In prestart they overlap and it looks bad)
- Antediluvian movement fix max_moves=0 not supported under [side], better use ai_special=guardian

Still in draft until I finish part II

### DIFF
--- a/scenarios6/01_The_Awakening.cfg
+++ b/scenarios6/01_The_Awakening.cfg
@@ -61,36 +61,6 @@
                 [/show_if]
             [/objective]
         [/objectives]
-        [message]
-            speaker=narrator
-            image="wesnoth-icon.png"
-            message= _ "In the dark, in a forgotten tomb deep under all places known to mankind..."
-        [/message]
-        [message]
-            speaker=narrator
-            image="wesnoth-icon.png"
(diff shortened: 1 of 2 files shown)


Dugy:
Some prestart messages were intended to be shown before the map is shown. Please take that into consideration.

Toranks:
If I see that the map does narration spoilers, I can convert them to story messages with a flat black background.
Anyway, most of the ones I've seen so far go right after the story screen (which has its own narration), so it doesn't matter that the narrator speaks during the beginning of the scenario.

Dugy:
All right.

Toranks:
I have recreated the conversation with Councilor Charles to show him on the map instead of on prestart. This is the result:

https://github.com/Dugy/Legend_of_the_Invincibles/assets/40789364/0add30b9-30a4-454c-b701-744ae1ed0a4a

Dugy:
That conversation between Charles and Lethalia was taking place in an imaginary palace conference room represented by a black background. Your change turns it into a ramparts chat. Though... if I got a picture of a palace conference room, it would not be able to use that as background in the prestart event. So I guess it's better your way.

Toranks:
It can also be moved to [story] [part] with the background you say. I did it this way because the conversation in prestart was quite linear with the one in start, and it didn't lose meaning if it was shown on the map.

Dugy:
Overlaying portraits in [story] is a bit of a problem, let's keep it the way you've done it.

Toranks:
Conditionals also work under [story]
imagen

PR #514, opened 2023-06-24
"Into the light" scenario - soul eater variable error

Toranks:
$eater is the variable that uses the soul eater event and is deleted each time, you have to use the variable saved in the unit.

Do you prefer this type of small fixes in different PRs like you told me, or do you prefer me to join them in Part II fixes?

### DIFF
--- a/scenarios6/03_Into_the_Light.cfg
+++ b/scenarios6/03_Into_the_Light.cfg
@@ -212,7 +212,7 @@
         [unstore_unit]
             variable=lethalia
             {COLOR_HEAL}
-            text="$eater.variables.devour_count|/$eater.variables.max_devour_count"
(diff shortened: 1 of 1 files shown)


Dugy:
I prefer it this way, but it might be annoying to you.

If you want, you can make a larger PR, but squash closely related commits together, so that I could merge it as separate commits instead of squashing the PR into a single commit as it's usually done.

Toranks:
Actually, for me it's pretty much the same job one way or the other, using github desktop, so I'll submit most fixes as individual PRs, and use the Part II fix only in case of potential conflicts.

Dugy:
Okay, if individual PRs are fine for you, then let's keep it that way.

PR #515, opened 2023-06-25
Deprecated

white-haired-uncle:
tested where possible, though in a couple cases I can't find anything that would actually trigger the piece of code in question

### DIFF
--- a/lua/inventory/items.lua
+++ b/lua/inventory/items.lua
@@ -34,7 +34,7 @@ local inventory_config = {
 				-- This may be a unit on recall list, so make sure that "unit" variable
 				-- (which is needed in unit_information_part_1()) gets populated anyway.
 				wml.variables["unit"] = unit.__cfg
-				wesnoth.fire_event("unit information", unit)
+				wesnoth.game_events.fire("unit information", unit)
 				wml.variables["unit"] = nil
 			end,
 		},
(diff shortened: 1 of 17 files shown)


Dugy:
Where are these messages about deprecation shown?

white-haired-uncle:
They show up on the console/log with debug enabled (I was using wesnoth -debug --log-debug=scripting/lua --no-log-to-file ).  Playing the tutorial is a simple way to trigger them.

Dugy:
Aha... good to know.

Toranks:
Don't update wesnoth.fire_event with wesnoth.game_events.fire in 1.16, or it won't work. It is a function of 1.17. The old version must be kept. The deprecated error only shows up in debug mode, a normal player won't see it, but with these changes a normal player in 1.16 will see a error of invalid function and it won't work.

Toranks:
Don't update wesnoth.fire_event with wesnoth.game_events.fire in 1.16, or it won't work. It is a function of 1.17.6 and later only. The old version must be kept. The deprecated error only shows up in debug mode, a normal player won't see it, but with these changes a normal player in 1.16 will see the error and it won't work.

https://wiki.wesnoth.org/LuaAPI/wesnoth/game_events

white-haired-uncle:
Ugh.  Good catch.  My test box defaulted to 1.17.

PR #518, opened 2023-06-26
{FOREACH} => [foreach]/[for]

white-haired-uncle:
08:09:27 info deprecation: FOREACH has been deprecated indefinitely.; You should use the [foreach] WML tag instead.

These have all been tested to at least perform equivalently to the previous versions (I don't think 03 works as intended, but not due to these changes, see https://github.com/Dugy/Legend_of_the_Invincibles/issues/517).

### DIFF
--- a/scenarios6/03_Into_the_Light.cfg
+++ b/scenarios6/03_Into_the_Light.cfg
@@ -102,24 +102,27 @@
         [filter]
             side=2
         [/filter]
-        {FOREACH unit.modifications.trait i}
-            [if]
-                [variable]
-                    name=unit.modifications.trait[$i].id
-                    equals=slimy
-                [/variable]
-                [or]
+        [for]
+	    array=unit.modifications.trait 
+	    [do]
+                [if]
                     [variable]
                         name=unit.modifications.trait[$i].id
-                        equals=regrowing
+                        equals=slimy
                     [/variable]
-                [/or]
-                [then]
-                    {CLEAR_VARIABLE unit.modifications.trait[$i]}
-                    {VARIABLE_OP i sub 1}
-                [/then]
-            [/if]
-        {NEXT i}
+                    [or]
+                        [variable]
+                            name=unit.modifications.trait[$i].id
+                            equals=regrowing
+                        [/variable]
+                    [/or]
+                    [then]
+                        {CLEAR_VARIABLE unit.modifications.trait[$i]}
+                        {VARIABLE_OP i sub 1}
+                    [/then]
+                [/if]
+	    [/do]
+	[/for]
         [unstore_unit]
             variable=unit
             find_vacant=no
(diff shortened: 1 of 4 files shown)


Toranks:
This part has an error:

```
                [foreach]
                    [teleport]
		    array=teleport_store
                        [filter]
		    [do]
                            id=$teleport_store[$i].id
                        [teleport]
                        [/filter]
                            [filter]
                        x,y=$x1,$y1
                                id=$teleport_store[$i].id
                        clear_shroud=yes
                            [/filter]
                        animate=no
                            x,y=$x1,$y1
                    [/teleport]
                            clear_shroud=yes
                {NEXT i}
                            animate=no
                        [/teleport]
		    [/do]
                [/foreach]
```

You must change $teleport_store[$i] with `$this_item`, so `id=$this_item.id`

Toranks:
This part has an error:

```
                [foreach]
		    array=teleport_store
		    [do]
                        [teleport]
                            [filter]
                                id=$teleport_store[$i].id
                            [/filter]
                            x,y=$x1,$y1
                            clear_shroud=yes
                            animate=no
                        [/teleport]
		    [/do]
                [/foreach]
```

`id=$teleport_store[$i].id` must be `id=$this_item.id`

white-haired-uncle:
Hmm, I'm not seeing the difference.  Since i is the default loop variable, aren't $teleport_store[$i] and $this_item just different names for the same thing?

All of my units did get teleported then I tested this.

white-haired-uncle:
I have been watching for loops that make changes to the loop variable/array.  And not changing anything I don't have to.

But I can see where it might make sense to use this_item "everywhere", if you have to use it in some cases anyway.  And it's easier to read, IMO.  Think it's better that way?

thanks

Toranks:
For me it is easier to read because when I started learning WML FOREACH macro it was already deprecated and I used always [foreach]. I do not know in the case of someone who has been around for longer.

Dugy:
Well, this would mean that one has to be careful when using `[foreach]` and deleting something from the array, because the decrementation of `i` at the end of the loop will be ignored and the next element will be skipped.

Toranks:
Yes, in fact I don't recommend using foreach when manipulating the variables inside that involves deletion of elements. It happened to me before and I was on discord trying to solve the cause of the error. 
In this case it's a teleport, so it's not relevant.

white-haired-uncle:
I've been using for when there is any modification to the loop variable/array (such as removing traits).

This assumes me...

Deprecation message:
info deprecation: FOREACH has been deprecated indefinitely.; You should use the [foreach] WML tag instead.

wiki description for [foreach]:
Loops over an array variable. This is not used to implement the {FOREACH} macro because [foreach] doesn't support adding or removing elements while looping (and will usually raise an error), use [for] instead.

PR #519, opened 2023-06-26
Into the light

white-haired-uncle:
Fixes logical error removing trait slimy.

Moves said code to utils/abilities.cfg
  1) Now that I think of it, that may not be the right place (I was originally trying to remove abilities, not traits)
  2) I'm not entirely sure if "passing" unit makes sense.  Not sure if unit is a reserved word, default variable or what.  What I have is tested and working, but.
  3) I think this could be used in other scenarios.  
  4) I don't see any other areas where only a single trait is removed, so moving this won't break consistency.
  3) I have a copy of this scenario that just fixes the slimy issue with the remove code in place if you prefer.

Code works, better now, but still doesn't do what I think it was intended to.  Ref: https://github.com/Dugy/Legend_of_the_Invincibles/issues/517

### DIFF
--- a/scenarios6/03_Into_the_Light.cfg
+++ b/scenarios6/03_Into_the_Light.cfg
@@ -102,27 +102,11 @@
         [filter]
             side=2
         [/filter]
-        [for]
-	    array=unit.modifications.trait 
-	    [do]
-                [if]
-                    [variable]
-                        name=unit.modifications.trait[$i].id
-                        equals=slimy
-                    [/variable]
-                    [or]
-                        [variable]
-                            name=unit.modifications.trait[$i].id
-                            equals=regrowing
-                        [/variable]
-                    [/or]
-                    [then]
-                        {CLEAR_VARIABLE unit.modifications.trait[$i]}
-                        {VARIABLE_OP i sub 1}
-                    [/then]
-                [/if]
-	    [/do]
-	[/for]
+        #
+        # Removing the trait doesn't actually remove the effects, so this isn't accomplishing much
+        #
+        {REMOVE_TRAIT_BY_ID unit "regrowing"}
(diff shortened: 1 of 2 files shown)


Dugy:
> Removing the trait doesn't actually remove the effects, so this isn't accomplishing much

I think I did this very wrong. The traits' effects can be removed by using [transform_unit] to change it to the same type, it will reconstruct the unit's properties from its modifications.

Toranks:
There is a error here:

```
#define REMOVE_TRAIT_BY_ID UNIT TRAIT_ID
    [for]
        array={UNIT}.modifications.trait
        [do]
            [variable]
                name={UNIT}.modifications.trait[$i].id
                equals={TRAIT_ID}
            [/variable]
            [then]
                {CLEAR_VARIABLE {UNIT}.modifications.trait[$i]}
                {VARIABLE_OP i sub 1}
            [/then]
        [/if]
    [/do]
[/for]
#enddef
```

Must be:

```
#define REMOVE_TRAIT_BY_ID UNIT TRAIT_ID
    [for]
        array={UNIT}.modifications.trait
        [do]
            [if]
                [variable]
                    name={UNIT}.modifications.trait[$i].id
                    equals={TRAIT_ID}
                [/variable]
                [then]
                    {CLEAR_VARIABLE {UNIT}.modifications.trait[$i]}
                    {VARIABLE_OP i sub 1}
                [/then]
            [/if]
        [/do]
    [/for]
#enddef
```

Toranks:
It may also be necessary
```
     [update_stats]
         id=$unit.id
     [/update_stats]
```
Just after transform unit

white-haired-uncle:
Yes.  I thought I fixed that.  I was testing on another machine and must have fixed it in the wrong place.  Thanks.

PR #522, opened 2023-06-27
Disable undo on DISABLE_UPKEEP macro to avoid desynchronizations

Toranks:
If you undo and redo a recall/recruit with this macro enabled, the game will fail and prompt you to save and exit. You can continue playing normally if you ignore it, but it's better to avoid the error.


Discontinuum:
Strange that it allows undo, I thought having an event without explicit "[allow_undo]" would stop undoing

PR #523, opened 2023-06-28
{FOREACH} => [foreach]/[for]

white-haired-uncle:
08 format is odd.  I think it's the EARTHQUAKE.

Looks like Lethalia_evasion event and transport units to leader could probably be made into macros.

Did a little cleanup on evasion and eruption.

I don't understand the Lethstore stuff in 12.  Since the {VARIABLE moved 0} is inside the FOREACH/for loop, isn't it always going to be 0 when evaluated?  Guessing this is a typo, and we could use break instead?

Tested one instance of each of the events (on 1.16 this time).

### DIFF
--- a/12_The_Last_Enemy.cfg
+++ b/12_The_Last_Enemy.cfg
@@ -1,0 +1,595 @@
+#textdomain wesnoth-loti
+#define LETHALIA_BLEED TYPE X Y
+    [unit]
+        type=Mist
+        x,y={X},{Y}
+        side=2
+        id=bleed_mist
+    [/unit]
+    {ADVANCE_UNIT type=Mist "{TYPE}"}
+    [store_unit]
+        [filter]
+            id=bleed_mist
+        [/filter]
+        kill=yes
+        variable=dev_null
+    [/store_unit]
+    {GENERIC_UNIT 2 {TYPE} $dev_null.x $dev_null.y}
+    {CLEAR_VARIABLE dev_null}
+#enddef
+[scenario]
+    id="12_The_Last_Enemy"
+    name= _ "The Last Enemy"
+    map_data="{~add-ons/Legend_of_the_Invincibles/maps/12_Escape_Exit.map}"
+    next_scenario=13_Avengers
+    victory_when_enemies_defeated=no
+    turns=-1
+    experience_modifier=80
+    {GLOBAL_EVENTS}
+    {SCENARIO_MUSIC "battle_epic.ogg"}
+    {EXTRA_SCENARIO_MUSIC battle.ogg}
+    {EXTRA_SCENARIO_MUSIC weight_of_revenge.ogg}
+    [time]
+        id=dark_indoors
+        name= _ "Indoors"
+        image=misc/time-schedules/schedule-indoors.png
+        lawful_bonus=-25
+        red=-45
+        green=-35
+        blue=-10
+    [/time]
+    [time_area]
+        x=42-45
+        y=28-34
+        {TWO_SUNS_DEFAULT_SCHEDULE}
+    [/time_area]
+    [time_area]
+        x=39-41
+        y=33-34
+        {TWO_SUNS_DEFAULT_SCHEDULE}
+    [/time_area]
+    [side]
+        type=Efraim_god
+        id=Efraim
+        name=_"Efraim"
+        canrecruit=yes
+        side=1
+        controller=human
+        recruit=Spearman,Swordsman,Javelineer,Pikeman,Bowman,Longbowman,Heavy Infantryman,Shock Trooper,Fencer,Duelist,Mage,White Mage,Red Mage,Cavalryman,Dragoon,Horseman,Lancer,Snow Hunter,Knight,Elvish Fighter,Elvish Archer,Elvish Shaman,Elvish Scout,Elvish Druid,Elvish Sorceress,Elvish Captain,Elvish Hero,Elvish Marksman,Elvish Ranger,Elvish Marshal,Elvish Champion,Elvish Sharpshooter,Elvish Avenger,Elvish Outrider,Elvish Shyde,Elvish Enchantress,Dwarvish Fighter,Dwarvish Steelclad,Dwarvish Guardsman,Dwarvish Stalwart,Dwarvish Thunderer,Dwarvish Thunderguard,Dwarvish Ulfserker,Dwarvish Berserker
+        gold=300
+        village_gold=2
+        team_name=good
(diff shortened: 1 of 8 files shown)


white-haired-uncle:
Hopefully I didn't break translation.  Don't know how that works.

Dugy:
You broke the translation. Any change to a translatable string breaks it.

white-haired-uncle:
Bummer.  I yanked that part out.

So, I assume any changes to translatable strings will require translators to make a change and therefore should be considered carefully.  For example, IMO that message really should say "24 fire damage", not "24 damage".  But, if that one little change causes others to have to update their translations, there's a question of whether the change is really worth making.  And putting variables in translatable strings is pretty much never going to be worth it, as any change to the variables would mean the translations may need to be updated (say from singular to plural), but without this being obvious to the translators and could even happen in the middle of a game where which would just mess things up completely.

Dugy:
True, changing in a translatable string has a 100% chance of requiring changes to translation, changing the value of a variable invoked inside a translatable string has a lower than 100% probability of requiring changes to translation because of changes in inflection.

Though, I'd prefer replacing numbers with variables in translatable strings only when the change actually happens, so that the change is not done before it's necessary.

Do you want me to merge this now?

Dugy:
I think the code may have deteriorated over time. The clones were supposed to have the same health and experience, and she was supposed to be swapped with one of those newly created clones. Because the order of units stored is rather unpredictable, swapping her with the first one should do the trick too, but it's not the way it was intended.

As far as I know, it's possible to hold the flood of clones by recruiting a lot of Level 1 Mages (who have a good chance to hit an elusive enemy), then surround the real one with stronger units before starting to attack - she cannot sidestep when she has no free hexes around her. Or did this break too?

white-haired-uncle:
The way it was, it was only checking one unit.  If that one had the same amount of HP, she swapped with it.  Doesn't mean it was one of the latest batch to be cloned, though.  If the HP doesn't match, she doesn't swap with anyone.

I changed it to loop over the clones, fixed the move variable, so now it looks until it finds a match and she always swaps position with one of her clones (the first one we find with same HP).  It's not really random, at least for the first few turns I can always tell where she is since it's always the same (turn 1 she's at 18/7, turn 2 she's at 23/20, etc.  Easy enough to fix by just picking a random number for the index, repeat until you find an HP match.  But I don't know that it really matters.

I tried the mage approach.  I hadn't thought of that.  It seemed promising at first, but I never could get it to work.  She is creating clones at right about the same rate you can recruit mages (assuming no one makes it into your castle).  If your mage misses a clone, it runs away.  If a clone misses your mage, it gets to try again, eventually killing your mage and usually at least one other (she's got cleave).  And once you hit a bad run of the dice where your mages miss a few in a row, you're really toast.  Even with great luck, I don't see when you'd have time to recall the heavy hitters you need to put her down.

I did manage to get her trapped between a couple walls once so she couldn't run away.  But then on the next turn, she swapped places with one of her clones, and even knowing where she was I couldn't get to her.

I actually think "fixing" the code to work as intended ended up making it much harder (like nearly impossible).  The units in lethstore.reflections are pretty much in the order in which they were created.  If you only look at lethstore.reflections[0], as soon as you hit her and drop her HP, it's unlikely she'll find a match to swap with (looking at the last entry would probably almost guarantee a match) so once you identify her she probably won't move.

Dugy:
Apparently, there are quite some bugs there now.

Regarding difficulty, which difficulty are you playing? If was usually balancing the scenarios for normal difficulty, easy was just tweaked a bit to be easier, hard to tweaked a bit to be harder, without a guarantee that hard is playable and easy is not too easy. Because I can't really tell if something is playable when I am not one of the best players. So if you think it's too hard, it shouldn't be a problem to reduce the number of clones instead of allowing bugs to reveal which one is the real one.

white-haired-uncle:
Okay, I made the swap unpredictable, even to the point of having her remain in place sometimes.  Tested a few times to ensure she doesn't follow a noticable pattern.

I think I'm done with chapter 7 now.

The Last Enemy is probably still too hard, but I hate to mess with it based only on my opinion.  I asked on the forum, and I'm sure someone will complain if it's too tricky.  I was able to beat it once using mages, but probably only because I had a turn where every mage hit and I was able to identify her and she was within range that I could hit her with the big guns before she could escape.

I'm testing on normal.  Generally, I don't find much difference in difficulty levels, as making one scenario more challenging usually just means your units advance more.

Dugy:
Cool.

Btw, hard difficulty doesn't just increase the number of enemies, it does other changes that won't be compensated by more experience gained, such as some enemy types having more damage and hitpoints, items being weaker or the random upgraded units appearing more often.

white-haired-uncle:
Just noticed I put 12_The_Last_Enemy.cfg into the root, not scenarios7.

Sorry, I don't know how to fix that.

Dugy:
If you have access to git's command line interface, you can do this:
```
git rm scenarios7/12_The_Last_Enemy.cfg 
git mv 12_The_Last_Enemy.cfg scenarios7/12_The_Last_Enemy.cfg
```
However, I am getting invalid closing tag when I pull the latest master, I am not sure what's going on, so I am not doing this myself.

white-haired-uncle:
Thanks.  Using the cli has been on my todo list, and this gave me the 
excuse to get started.  I don't care much for the web method of 
uploading files, too easy to mess and up no way I've found to correct 
w/o starting over.

Now I just need to learn to remove commits, and hopefully run the 
luacheck without having to create a pull request first (I always end up 
with at least one whitespace character at the end of the line 
somewhere).

Dugy:
I think you may have uploaded something wrong, probably using tabs instead of spaces without having set up git filter? It's causing me trouble when using git filter. Can you please fix it by running wmllint and committing the changes? It's very difficult for me to do when I am using git filter.

white-haired-uncle:
Sure. Will take me a bit while I try to figure out how to download all 
files in a pr (this github cli thing?).

I'll look into git filter.  I try to run wmlindent on everything, but I 
may have missed something.  Or it's the file 08_Eruption.cfg, which 
wmlindent did not like.

white-haired-uncle:
I'm wondering about the EARTHQUAKE line, it looks strange but I don't 
understand it yet.

Dugy:
The command to allow switching to PR branches, you have to use `git remote add` and the address of the fork from where you're taking the PR.

Earthquake is fine, I am having problem with scenarios Tutorial, Friends No More, Across the Barren Land and Frozen Land.

PR #525, opened 2023-06-29
Improved devastating blow

Toranks:
On the one hand, via lua a name and description are also inserted so that it is visible in the weapons in which it is applied.

On the other hand, the event is changed to be a real hit with [harm_unit], made by the attacker, with defense animation, for the same value (20% of the victim's HP). Previously it was an invisible hit and if the unit died it would suddenly disappear with no animation.

imagen

### DIFF
--- a/lua/stats.lua
+++ b/lua/stats.lua
@@ -291,7 +291,7 @@ function wesnoth.update_stats(original)
 				weap.description = bonuses.name
 			end
 			if bonuses.devastating_blow > 0 then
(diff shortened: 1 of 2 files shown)


Toranks:
Done. What you do think? This is a simulation with 100% of chance of devastating blow:

https://github.com/Dugy/Legend_of_the_Invincibles/assets/40789364/18e59052-4b39-4f2a-aa35-ca04c79aeb27

- harm_unit acts silently, without animation and without killing the unit.
- Float text appears first, and a 300ms delay is left to avoid overlapping text.
- Integrated translation instead of concatenation.

Dugy:
Fine, cool, thanks.

Toranks:
Oh, by the way, you can add a _ in `text=_"20% HP"` ? I forgot

Dugy:
Yeah, that could be useful. And minus 20% HP.

PR #526, opened 2023-06-29
Lua revert 1.17 specific changes

white-haired-uncle:
tested on 1.16


Dugy:
Are you sure you only wanted to add one file?

white-haired-uncle:
I added two.  luacheck didn't like one, so I fixed it and uploaded a new version.  Now I'm trying to figure out how to make the check run again.

Seems to me like deleting the offending commit would be a way to go, but I haven't found that possible.

Toranks:
Why items.lua is duplicated on the main folder?

white-haired-uncle:
I did fix the whitespace, but I don't know what to do after that.  Besides just delete and start over.

Toranks:
That whitespace is still present in the last commit. Add another with that removed, and luacheck will pass in a few minutes.

white-haired-uncle:
Thank you, I finally found a way to edit that file w/o uploading a new one.

Now I just have to remove the items.lua in the root directory.  I messed this one up good.

PR #527, opened 2023-06-30
RARE_ITEM -> INVENTORY_ITEM

white-haired-uncle:
I don't have a save to test this against, but I think it's just a simple oversight.

I went to Nightmare Cellar and left on turn 1.  All of the loot was automatically picked up for me.  Since every other scenario in ch9 uses INF_ITEM, I bet this is the fix.

### DIFF
--- a/scenarios9/30_Nightmare_Cellar.cfg
+++ b/scenarios9/30_Nightmare_Cellar.cfg
@@ -69,9 +69,9 @@
     [event]
         name=start
         {INFERNO_RECALL_ALL}
-        {RARE_ITEM 23 5}
-        {RARE_ITEM 36 6}
-        {RARE_ITEM 28 8}
+	{INFERNO_ITEM 23 5 sc30_item1}
+	{INFERNO_ITEM 36 6 sc30_item2}
+	{INFERNO_ITEM 28 8 sc30_item3}
         [if]
             [variable]
                 name=quests.entered_sc30


Dugy:
That would not change anything, INFERNO_ITEM guarantees that the item will not be there again when you return into the area, and may have some changed item type distribution.

The item was picked because it was dropped on ground the same turn as you ended the scenario, which is a mechanism for ensuring you don't lose an item dropped by the last leader in a _defeat all enemy leaders_ scenario. It should not work on items on the ground, but it's possible the conditions also match on ground items if you finish the scenario in turn 1.

white-haired-uncle:
Well that's too bad, it seemed like the obvious difference, since I only 
saw this on NC and it's the only one that uses RARE_ITEM.

Something's up, because I don't think anything I did should have 
triggered it, like triggering victory (my first thought was that 
scenario was configured with result=victory just for leaving).  I just 
popped in to juggle some items around between units on the autorecall 
list, then went back to the previous scenario.

I'll look at it again when I get there in the game.  I had multiple 
issues with this scenario last time (free items, Lethalia_evil supposed 
to be part of a discussion even though she doesn't exist, and Lethalia 
making a comment related to Abaddon/Lillith/Uria appearing when they 
didn't).

white-haired-uncle:
"INFERNO_ITEM guarantees that the item will not be there again when you return into the area"

So if we leave it using RARE_ITEM, does that mean I can keep going back to loot those locations?

Dugy:
To solve the autopickup exploit, it's probably necessary to add an optional argument to the lua code that generates the item that would tell it that the item was created at the start of the scenario and is not subject to end of scenario autopickup. This can be achieved by setting the variable indicating the turn when the item was dropped to -9000, so that it would never be equal to the turn when you can end the scenario.

white-haired-uncle:
How about:

# We want player to be able to return and earn more loot, so forget history in case they've already been picked up
{CLEAR_VARIABLE sc30_item1,sc30_item2,sc30_item3}
{INFERNO_ITEM 23 5 sc30_item1}
{INFERNO_ITEM 36 6 sc30_item2}
{INFERNO_ITEM 28 8 sc30_item3}

Dugy:
The RARE_ITEM macro has a different drop distribution than INFERNO_ITEM.

white-haired-uncle:
I was afraid of that.  What's that saying about "simple, elegant, and wrong"?

I'll look into the turns approach when I get there in the game. I was also wondering about victory conditions, I think I saw a macro which triggers the item pick up on victory -- though I have a feeling that will be a dead end.

Dugy:
Well, INFERNO_ITEM would not prevent the item autocollect from putting the items into your inventory if you finish in the first turn.

PR #529, opened 2023-06-30
Fix pr523

white-haired-uncle:
I changed the EARTHQUAKE line in Eruption, and now wmlindent doesn't complain about it.  It used to say 

wmlindent: "08_Eruption.cfg", line 439: close tag [/scenario] with indent already zero.

It still looks odd to me, and I haven't tried to figure out what that line is up to (my guess is it's a typo and there's a couple redraws that don't belong?).  

No changes to the rest of the files except running wmlindent.

### DIFF
--- a/scenarios1/00_Tutorial.cfg
+++ b/scenarios1/00_Tutorial.cfg
@@ -586,8 +586,8 @@ Take this item?"
                                         delayed_variable_substitution=no
                                         {VARIABLE tutorial.proceed 0}
                                         [remove_event]
-					    id=dont_take_it
-					[/remove_event]
+                                            id=dont_take_it
+                                        [/remove_event]
                                         [while]
                                             [variable]
                                                 name=tutorial.proceed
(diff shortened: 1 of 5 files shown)


Dugy:
Yeah, it's all right now. I don't know what is wrong with Eruption at the moment, some opening or closing tag may be in a macro or something.

white-haired-uncle:
Like this?

   {EARTHQUAKE (           [redraw] [/redraw])}

(There was a newline between the redraws before I got to it)

Dugy:
Maybe, I don't know.

What I know that causes problems is:
```
#define START_SOMETHING
[something]
  blablabla=yes
#enddef

{START_SOMETHING}
  blaaaabla=no
[/something]
```

But there may be other constructs that cause the problems. The usual way to debug it is to delete most events, see if it's fine, then either delete more events until it passes or try to delete less and less events if it's not passing.

PR #530, opened 2023-06-30
Devastating blow fixes

Toranks:
Fixes some things of #525 

- if a unit receives devastating blow while already dead, since the operation involves a variable already gone, this error is thrown, so I add a conditional so that units with 1 or less health do not receive this blow.
imagen

- Added in the description "It cannot kill"
- "-20% HP" translatable
- Wrong math operation. The previous subtracts 17%, not 20%


Dugy:
Wait, doesn't this revive a unit that would be killed by the hit in case the devastating blow didn't trigger?

PR #533, opened 2023-07-01
Offensively weapons descriptions

Toranks:
Despite the fact that many weapons indicate whether their use can be offensive or defensive, these do not have such a clarification. I include these using the same style as the others. Seem to you? Translating I realized that there were many like this.
If anyone knows of any other, I can include.

Note: In parry "When a defender target misses" means that if the defender of YOUR attack misses, so offensively only. I hope it is well understood.


Dugy:
Yes, I admit there were some inaccuracies.

PR #535, opened 2023-07-01
Foreach ch8

white-haired-uncle:
Replaced {FOREACH}

Removed some unused code

Note:  In Jungle Hell, around line 435, I'm assuming there should be only one match, so only the first match is selected.  The existing code would select the last one it found.

Note:  In United, all those upgrades wouldn't happen to be in an array somewhere that we could loop over instead of hardcoding so many if blocks?


Dugy:
> Note: In United, all those upgrades wouldn't happen to be in an array somewhere that we could loop over instead of hardcoding so many if blocks?

The redeem implementation changed a lot since I wrote that part and it's some leftover from the old times when it was clunky. There is no reason any more for it to be written like this.

PR #537, opened 2023-07-01
use new wml sqrt

white-haired-uncle:
requires >= 1.15.10

tested on 1.16.2

code just below this uses old distance result to set limits, so might need to tweak

I needed a distraction and ran into this comment while looking for something else.

There's also an abs now, fwiw


Dugy:
Let's see if something will start behaving slightly oddly.

Issue #538, opened 2023-07-01
Knockback on Solid Phoenix kills adjacent allied unit


Toranks:
I don't know why it happens only with Solid Phoenix. I've used debug to swap the Solid Phoenix for another unit, and the bug doesn't occur.

How to reproduce:
- Attack the Phoenix with the knockback attack of the Lunatic Knight
- The witch in the back dissapears and the phoenix takes its place

[Knockback kills ally.gz](https://github.com/Dugy/Legend_of_the_Invincibles/files/11927297/Knockback.kills.ally.gz)

EDIT: Found the cause.

NOTE: There is a lua error that says `[unstore_unit] variable second_unit does not exist`. This bug is localized and belongs to the scenario (the event that disables fire attacks). I'll try to send a PR to fix it soon, but it has nothing to do with this bug. EDIT: DONE

Toranks:
FOUND THE CAUSE: I'ts because Solid Phoenix has all terrains non-passable, so find_vacant and check_passability works bad. Reported on #modding waiting for them to tell me if it is a bug or if the code needs to be modified.

Toranks:
@Dugy @Discontinuum  
They have already explained to me how to fix it. There are two alternatives for the solution. What behavior do you want immobile units to have if knockback is used on them?:
- The unit is moved with check_passability=no to the closest free hex to the destination one.
- The unit does not move.

Dugy:
I think it would be better if the unit didn't move and knowckback did nothing.

Toranks:
That has the simplest solution

Issue #385, opened 2021-11-25
If you add "meteor" or "lava flow" with a great spell but don't use these attacks before the end of the scenario, your leader stats wiped before the next recalc


Discontinuum:
Similar to #363 but in a different version: the object is added with "scenario" duration. If you use the attack, it's deleted by LotI's code and all is okay. But if you quit the scenario, I guess the game calls [remove_object] under the hood to wipe "scenario"-long objects which causes the same erasure of stats as in #363

Dugy:
Maybe the removal could be an end scenario event force triggered by using the attack.

Discontinuum:
Or make named events for deleting attacks and call them both during attacking and during exiting a level

PR #539, opened 2023-07-02
Better removal of fire attacks on Frozen Land

Toranks:
Instead of a complicated event system, the effect of which is that if you try to attack with a fire attack, the attack does not occur and the unit loses all its attacks instantly.
With this method, the fire attacks disappear for this scenario and that's it.

Also resolves a problem described on #538 with an ocasional lua error (not related with the knockback of the Solid Phoenix)

### DIFF
--- a/scenarios6/14_Frozen_Land.cfg
+++ b/scenarios6/14_Frozen_Land.cfg
@@ -242,78 +242,20 @@
 
     #Disable fire attacks
     [event]
-        name=attack
-        first_time_only=no
-        [for]
-            array=unit.attack
-            [do]
-                {VARIABLE unit.variables.former_attack_numbers[$i].number $unit.attack[$i].number}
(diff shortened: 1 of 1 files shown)


Discontinuum:
Note that using scenario-long objects will cause reset of LotI items effects on all affected units if you don't call update_stats again (for example https://github.com/Dugy/Legend_of_the_Invincibles/issues/385 )

Toranks:
I know, but the item is lost at the end of the scenario, and LOTI already has a way to do [update_stats] in the next scenario, when they are returned to the map (recalled or placed via events).

Toranks:
Oh, you are right, units do not reload their items. I'll add [update_stats] in the proper place.

Toranks:
Now works, tested when go to the next scenario

PR #540, opened 2023-07-02
Improved knockback

### LINKED ISSUE #538: Knockback on Solid Phoenix kills adjacent allied unit

Toranks:
- Change use of store/unstore to teleport
- check_passability works better inside [teleport] than in [unstore_unit], so it's more bug-proof
- Simplified code
- Immobile units cannot be knockbacked
- Since unstore_unit is not used, the unit dies on the spot if it has 1 or less health, as in the original code, so it can be animated without glitches.
- Resolves #538

### DIFF
--- a/utils/abilities_events.cfg
+++ b/utils/abilities_events.cfg
@@ -489,7 +489,7 @@
             [then]
                 [kill]
                     id=$second_unit.id
-                    animate=no
+                    animate=yes
                     experience=yes
                     fire_event=yes
                     [secondary_unit]
--- a/utils/abilities_events.cfg
+++ b/utils/abilities_events.cfg
@@ -486,62 +486,39 @@
                 name=second_unit.hitpoints
                 less_than_equal_to=1
             [/variable]
+            [then]
+                [kill]
+                    id=$second_unit.id
+                    animate=no
+                    experience=yes
+                    fire_event=yes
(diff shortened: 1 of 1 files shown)


Toranks:
This is also an improvement for many bosses who are locked in their keep and castle and the rest of the squares are impassable for him. I've seen it a lot throughout LOTI. Now they will not be able to be pushed out of their castles with knockback.

Toranks:
Amplificators still can be knockbacked, because normal terrains are passable for them, even if doesn't have enough moves for it (only 1). This isn't problem anyways, they can't be knockbacked to cliffs or something.

PR #541, opened 2023-07-02
Undo fire attack removal changes on Frozen Land

Toranks:
The new change using [object] gives an insurmountable problem:
If the unit advances or changes weapons to one that gives it a new fire attack, the object doesn't apply to this new attack even if the unit is rebuilt, since according to the developers, objects and advancements are applied in the strict order they were obtained.
So I think it's better to go back to the previous method, simply adding a conditional to prevent any error from appearing on the screen in case unit or second_unit are not present on attack end event.
This is the only change from the old code.

Remember it still has a small inconvenient, but it doesn't affect the balance much:
- The attacker loses all its attacks for trying.
- Both units receives experience as if it had attacked.

Undo #539

### DIFF
--- a/scenarios6/14_Frozen_Land.cfg
+++ b/scenarios6/14_Frozen_Land.cfg
@@ -291,10 +291,9 @@
             {CLEAR_VARIABLE unit.variables.former_attack_numbers}
 
             [if]
-                [variable]
-                    name=unit
-                    not_equals=$empty_variable
-                [/variable]
+                [have_unit]
+                    id=$unit.id
+                [/have_unit]
                 [then]
                     [unstore_unit]
                         variable=unit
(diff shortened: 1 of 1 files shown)


Toranks:
Trying other methods, I found this bug xD
https://github.com/wesnoth/wesnoth/issues/7760

Dugy:
If you want to make it perfect, it's possible to change `stats.lua` to add a scenario-scope post-recalculation event hook (similar to the `[on_equip]` property that some droppable items have) and use that to set the `attack_weight` of all fire attacks to 0.

Toranks:
After trying different alternatives, I haven't found a way to make it "perfect" with my current knowledge. It is better to revert until another way is found.

Dugy:
So you want me to accept this PR for now?

Toranks:
Yes, since this PR is basically a revert, you'll be able to work on it later with no change other than a conditional to avoid the lua bug I mentioned.

PR #542, opened 2023-07-03
Redeem count

white-haired-uncle:
Replaced a lot of wml with a little lua

lua function used is commented as temporary, added comment to point out where I used it

fixed error message typo in said function

tested in both scenarios, returns same results as original


white-haired-uncle:
Just noticed something else I want to test before this (might) gets merged

white-haired-uncle:
Okay.  Looking at lua/redeem.lua, I believe the intent was for that function to be called using something like

```
        [count_redeem_upgrades]
            to_variable=lethalia_upgrade_count
            find_in=     SOMETHING THAT PASSES "id=Lethalia"
        [/count_redeem_upgrades]

```
But, it looks like the find_in part never got implemented in the lua function.  I tried to implement it, and using [count_redeem_upgrades] does mostly work, but I no matter what I try to put in find_in, I get complaints that wesnoth.units.find_on_map wants to be passed a wml table but I can only seem to pass a string or a table, not a wml table.

But, calling that function works fine as implemented using [lua], so I'm okay with that (for now?).

BTW, I noticed ch8 United ends in victory if you get a lua error.  I don't know what the implications of that might be, but I don't think I've ever seen that happen before.

Dugy:
I don't remember having any issues with using filters in lua code, I don't know what could have happened there. But good that you got that working.

> BTW, I noticed ch8 United ends in victory if you get a lua error. I don't know what the implications of that might be, but I don't think I've ever seen that happen before.

Maybe the crash causes an enemy leader to despawn?

white-haired-uncle:
I can't tell, because I never get a chance to control the map view 
before the lua error and Efraim says "We have won" (or something like 
that), but I don't think the leaders ever get created.  At a wild guess, 
the lua error blows up the start/prestart, the game starts, sees no 
leader and victory_when_enemies_defeated=yes occurs?  Just a guess, I 
know almost nothing about that process.

Dugy:
Aha, yeah, that problem. The bosses were supposed to arrive at some point, a lua error aborts the execution of the event, so the bosses are not created and you win because enemies are defeated.

PR #543, opened 2023-07-03
Redeem count

white-haired-uncle:
I heard back from edwardspec, and he gave me a clue, find_in can be used with stored_units.

The new approach, using [count_redeem_upgrades], seems a little easier to read.

The first approach, using [lua], provides an example of using more flexible options, but I'm not sure when you would use them.


Dugy:
It's always useful to be more flexible if it's not costly.

PR #545, opened 2023-07-03
Clear Lethargy/Wrath and Redeem between scenarios

### LINKED ISSUE #544: Lethargy and wrath intensity is not cleared between scenarios
I am asking if this is intentional in the first place. The solution is simple, incorporating the same turn_refresh code into victory or start (or both).
This issue doesn't just affect units that have lethargy at the end of one scenario, stay lethargic the next scenario, and must wait multiple turns to heal. It also affects units that are saved to a variable and come back years later. This is the case of Efraim and Lethalia when they become gods (the one you have not chosen as the main one).

Toranks:
This is the simplest fix for now, as long as the method is not changed. Both are checked, and without using the {FOREACH} macro.

The old code uses direct edits to units variables to avoid rebuilding the unit, but it is still effective in the LOTI context and does not need to be updated, in my opinion. It works fine, and with this it's already treated as any status between scenarios. I play LOTI with various other addons and they mix well, so it's still a robust system even in 1.17

Resolves #544

### DIFF
--- a/utils/abilities_events.cfg
+++ b/utils/abilities_events.cfg
@@ -5568,6 +5568,7 @@
                 [/unstore_unit]
             [/do]
         [/for]
+        {CLEAR_VARIABLE wrathful}
     [/event]
 
     [event]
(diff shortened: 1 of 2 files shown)


Dugy:
This can do for now. Thanks.

PR #546, opened 2023-07-04
Revert redeeming reset on #545

Toranks:
To avoid an exploit in chapter 9


Dugy:
I thought you had made a mistake when it had 7 commits, but it does the right change when squashed :thinking:

Toranks:
That's because I used the same branch, and I had to update it first.

Dugy:
Ahaaaa... Now it makes sense.

PR #550, opened 2023-07-04
[effect] apply_to=level method on Beelzebub

### LINKED ISSUE #549: Beelzebub loses his level when he levels up
In Beelzebub's code, this operation is performed:

```
{VARIABLE beel_store_level_add {POWER}}
{VARIABLE_OP beel_store_level_add multiply 5}
{VARIABLE_OP beel_store.level add $beel_store_level_add}
[unstore_unit]
      variable=beel_store
      find_vacant=yes
[/unstore_unit]
```

Basically for the player to raise those experience points when fighting with him, and the corresponding when killing him.
The problem is that when he levels up, he go back to the base level of the Beelzebub unit type, which is 30.
Of course, it is something that happens rarely, but not so rare in the second part, where Lethalia and Efraim are each level 10.

The solution is easy in 1.17, using the new `[effect] apply_to=level`
Not so easy in 1.16, requires lua. But I have a working code for it, if you're interested.

Toranks:
- To enable this effect in versions of Wesnoth where it is not supported, I add the necessary code to _main.cfg along with loading the rest of the lua code. I do it this way so I can use the preprocessor. I don't know how to use the preprocessor within Lua, but if you know or prefer that I add it in a different place, tell me.

- When Beelzebub receives the buff objects, another object is added with [effect] apply_to=level set= with the value that was previously written directly to wml. Thus, the level will not be lost no matter what.

- For code efficiency, the status=non_living is changed to this object, instead of several repeated objects.

- The lua code has been used in a similar way for years by different addons, it is reliable.

- Tested in 1.16 and 1.17.

- Resolves #549


utils/utils.cfg
```diff
@@ -1289,13 +1289,26 @@ $item_info.description"
                                 to_variable=beel_store
                             [/unit]
                         )}
-                        {VARIABLE beel_store_level_add {POWER}}
-                        {VARIABLE_OP beel_store_level_add multiply 5}
-                        {VARIABLE_OP beel_store.level add $beel_store_level_add}
```
Dugy, line 1294: Please keep this for compatibility with 1.16.

_main.cfg
```diff
@@ -67,6 +67,26 @@
...
+                unit.level = cfg.set
+            elseif cfg.increase then
+                unit.level = unit.level + cfg.increase
+            else
+                wml.error("Invalid or missing key in [effect] apply_to=level")
+            end
+        end
+	>>
+[/lua]
+#endif
```
Dugy, line 89: Can you please put this in a separate file? I think this will be used much more once 1.18 goes out.

utils/utils.cfg
Toranks: apply_to=level works on 1.16 and 1.17 on the same way. The first with the new lua code, and the second natively.

_main.cfg
Toranks: Ok,  so another file on the lua folder

utils/utils.cfg
Dugy: Ahaa.... okay.

_main.cfg
Dugy: Yeah.

Dugy:
Can you please name that file `effects.lua`? The current file name is too vague.

Toranks:
I'm asking on discord about the lua error, but I'm reasonably sure it only applies to 1.17, not 1.16, since effects.level didn't exist in the last

Dugy:
I don't quite know why luacheck is failing. I think it's a false positive.

Toranks:
Yes, I have already tested it on both versions

Dugy:
It seems it's passing now.

Dugy:
Maybe there is a way to tell luacheck not to issue that warning using a comment or something?

Toranks:
I don't know how luacheck works, but Ravana tells me that it shouldn't mark effects.level as read-only, since other effects aren't read-only either. For now you can just ignore it?

Dugy:
I can't merge it with failing luacheck, it would keep failing on master and I would not know when an actual error appears. I don't know how you marked it as read only.

Toranks:
Well at the moment the only way I can think of is to revert the change and return the lua code to the [lua] tags in _main.cfg, where luackeck doesn't check.

Toranks:
The read only flag is a luacheck thing, not my code

Dugy:
According to https://stackoverflow.com/questions/49158143/how-to-ignore-luacheck-warnings you can suppress luacheck warnings:

```
-- luacheck: push ignore 121
foo() -- No warning.
-- luacheck: pop
foo() -- Warning is emitted.
```

Dugy:
Sorry, luacheck actually says it's warning 122, not 121.

Toranks:
Both give an error, it seems

Dugy:
No, now it's reporting a merge conflict.

Toranks:
> No, now it's reporting a merge conflict.

That was several minutes ago, I solved it in a few seconds

Dugy:
I can squash and merge it, though. That would not be a problem. I think something is wrong with the push, it reports an unpaired pop later in that file.

Toranks:
Leave it until tomorrow in case someone discovers something to fix this

Dugy:
Okay.

Toranks:
LOL sorry I wrote luacheck and luackeck respectively xDD

Dugy:
Lol, I thought you were copypasting it from my comment.

lua/effects.lua
```diff
@@ -0,0 +1,17 @@
+-- EFFECT TO CHANGE LEVEL
+-- Use on [effect]
+-- apply_to=level
+-- increase=number or $formula
+-- set=number or $formula
+
+-- luacheck: push ignore 122
+function wesnoth.effects.level(unit, cfg)
```
ProditorMagnus, line 8: Considering loti is one of the big ones and not expected to be compatible with other addons its not that relevant, but I assume using wesnoth.effects.loti_level would also solve the warning.

PR #552, opened 2023-07-06
typo in Apologies, update blizzard info in redeem

white-haired-uncle:
I swear blizzard used to scatter enemies, kind of like shockwave.  But it doesn't now, and I see no indication it's trying to.


Dugy:
I don't remember blizzard ever moving units around, maybe I wanted to do it but then I changed my mind and kept the original description there.

Issue #534, opened 2023-07-01
I haven't been using redeem


white-haired-uncle:
So I get to ch08s03 United.  Both E&L have redeem(3), and unit info shows they have each taken health and regeneration.  I take the upgrade and now he has health/regeneration/leadership, she has health/regeneration/heals, and both show redeem(5).

Looking at the code, it appears that if you have less than 5 redeem upgrades, and you choose to do so, it gives you enough to bring you up to 5 upgrades, but it just gives you the same upgrade over and over (for example, she probably has health/regeneration/heals/heals/heals).  Is the intended behaviour? 

If it is, and repeating heals multiple times is just intended to fill up those slots, then this is a horrible deal (especially if she already has heals, if that's possible), as there are tons of low level units to redeem in the next few scenarios.  But if it's not, and you're supposed to get enough real upgrades to get you to 5, it's too good a deal to believe.  Hence my confusion.  Based on the dialog, I think it's working as intended, but I had to read that over several times to come to that opinion.

And BTW, does the unstore need to be within the while loop?

I suspect Ch9,S1 has the same "issue".

thanks

white-haired-uncle:
Hmm.  Got to Ch9.  L had 4 upgrades (according to Unit Info), both redeem(4) and redeem(5), and was given the choice.  E had 5 upgrades, redeem(5) twice, and was not.  I don't remember if I took the upgrades in ch8 or not.  I wonder if I did, and they account for the double redeem values.  I also wonder if E should have been given the choice, or if something about taking the upgrade in ch8 with fake upgrades (leadership 3 times) is messing up the count when you get to ch9.

Dugy:
I think this part of the code was totally neglected for years. Redeem was changed to have a nice GUI, this was even backported to an ancient iOS version, but this part was overlooked, so it may look ugly.

However, this was not meant to compensate for negecting redeem. It was meant only to handle the problem that at some point, there might be so little level 1 enemies that redeem becomes nearly impossible to advance.

white-haired-uncle:
Okay, I think I understand what it's supposed to do (raise redeem level to 5 (or 7 for ch9) so you can redeem higher level enemies (and/or more at a time), but at the cost of not earning any more advancements until you have enough redeems that you would have reached level 5 (or 7) on your own).

If I got that, it's a horrible decision to take it.  Both of those scenarios have plenty of imps to practice on.

Also, what ever it's supposed to do, I don't think it's doing it right.

[...]

PR #553, opened 2023-07-06
fix United/Transporting redeem upgrade

white-haired-uncle:
This will break backward compatibility with saves that don't have unit.variables.redeem_level.  I could put in a workaround, if it's not available when needed (redeem advancement), derive it from the attack special.

I don't know why that other commit is getting included here.  Maybe I'm not keeping local branch up to date with remote?

Ref: #534

### DIFF
--- a/lua/redeem.lua
+++ b/lua/redeem.lua
@@ -36,7 +36,7 @@ known_ability_trees.redeem = {
 	},
 	blizzard = {
 		image = "attacks/blizzard.png",
-		label =_ "Blizzard (only for attacking, slows a lot of units, moves them over the place)",
+		label =_ "Blizzard (only for attacking, slows a lot of units)",
 		requires = { "arcticblast" },
 		short_name =_ "Blizzard"
 	},
--- a/lua/redeem.lua
+++ b/lua/redeem.lua
@@ -532,8 +532,6 @@ end
 -- Unit is identified by cfg.find_in parameter (e.g. find_in=secondary_unit).
 -- NOTE: this is TEMPORARY (won't be needed in the future),
 -- because the WML code that needs this variable might be replaced by Lua.
---
--- NOTE: this is now used by ch8 United and ch9 Transporting Facility.
 function wesnoth.wml_actions.count_redeem_upgrades(cfg)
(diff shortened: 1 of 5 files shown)


white-haired-uncle:
While looking at redeeming.cfg, I think I found a case where redeem_level might get reset to 1 on redeem advance.  I can't reproduce it, but I added a check to make sure it didn't happen.

Dugy:
Can you please make it count the upgrades in case the redeem count variable is nil so that there is some backwards compatibility?

Also, there is some lua code that is repeating in multiple scenarios without much change, can this be a a function in `redeeming.lua` that would just be called from scenario code?

If you don't figure out how to remove the unrelated commits from the feature branch, I can cherry-pick the correct ones and squash them.

white-haired-uncle:
I'm a little leery of counting upgrades because 
- the existing wml code and the existing lua code handle counting differently (absolute vs unique)
- there was an off-by-one issue with the wml code, at least in some places (N upgrades implies redeem (N+1), not redeem(N)).
- counting is dependent on lists of upgrades (currently separate lists) which could be error prone

I think the way to avoid these is just not depend on counting.

I can definitely put in a workaround for backward compatibility.  

That lua code does need to be centralized, and probably wrapped in wml tags.  But there's a lot I want to look further into in redeeming before I know exactly how (I think) it should look.

Dugy:
I know the counting is inconsistent, the goal here is just to make sure an old save will not be less usable than necessary because the redeem level suddenly comes down from some high value to 1.

Toranks:
How does the code finally work? Will you get the old redeem value as a variable to continue counting with the new method or will it reset to 0 but keeping advancements and skill level of redeem?

white-haired-uncle:
> How does the code finally work with previous saves? Will you get the old redeem value as a variable to continue counting with the new method or will it reset to 0 but keeping advancements and skill level of redeem?

When it checks for the unit.variables.redee.m_level, if it doesn't find anything it uses the old method to find the level, and sets the variable.

The level, counts, achievements are all preserved, there's just an easier way to get the level.

white-haired-uncle:
I forgot to close this. Superseded by #558

PR #559, opened 2023-07-09
Avoid desync problems with recall fresh troops

Toranks:
This event is triggered when a unit transfered from another side joins you the next scenario and you recall it, undo, then redo. It affects both multi and single player.
Gives exactly the same error as #522 

To avoid disabling undo on ALL recalls I've done a extra filtering (look at the three commits separately, ignore the last one, it's reverted as it was just a test)
- The allow_undo is removed
- Only the units that have the variables to be deleted or analyzed are filtered`*` (to avoid triggering any unit on side 1)
- Added another disable undo in unit placed event

`*` In case you don't know, `glob_on_variable=*` is a special key that works only when filtering wml (not on any tag), for example [filter_wml], and means "check for existence of `variable` with any value (`*`)"

### DIFF
--- a/utils/global_events.cfg
+++ b/utils/global_events.cfg
@@ -537,6 +537,22 @@
         first_time_only=no
         [filter]
             side=1
+            [and]
+                [not]
+                    [filter_wml]
+                        [variables]
+                            updated=yes
+                        [/variables]
+                    [/filter_wml]
+                [/not]
+                [or]
+                    [filter_wml]
+                        [variables]
+                            glob_on_starving=*
+                        [/variables]
+                    [/filter_wml]
+                [/or]
+            [/and]
         [/filter]
         [store_unit]
             [filter]
--- a/utils/global_events.cfg
+++ b/utils/global_events.cfg
@@ -529,6 +529,9 @@
                 [/not]
             [/filter]
             {UPDATE_STATS}
+            #Make the event undoable
+            {VARIABLE_OP disallow_undo rand(0..1)}
+            {CLEAR_VARIABLE disallow_undo}
         [/event]
      [/event]
      
--- a/utils/global_events.cfg
+++ b/utils/global_events.cfg
@@ -540,22 +540,11 @@
         first_time_only=no
         [filter]
             side=1
-            [and]
-                [not]
-                    [filter_wml]
-                        [variables]
-                            updated=yes
-                        [/variables]
-                    [/filter_wml]
-                [/not]
-                [or]
-                    [filter_wml]
-                        [variables]
-                            glob_on_starving=*
-                        [/variables]
-                    [/filter_wml]
-                [/or]
-            [/and]
+            [filter_wml]
+                [variables]
+                    glob_on_starving=*
+                [/variables]
+            [/filter_wml]
         [/filter]
         [store_unit]
             [filter]
@@ -564,15 +553,6 @@
             variable=advanced
             kill=no
         [/store_unit]
-        [if]
-            [variable]
-                name=advanced.variables.updated
-                equals=yes
-            [/variable]
-            [else]
-                {UPDATE_STATS}
-            [/else]
-        [/if]
         {CLEAR_VARIABLE advanced.variables.starving}
         [unstore_unit]
             variable=advanced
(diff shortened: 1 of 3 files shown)


utils/global_events.cfg
```diff
@@ -529,6 +529,9 @@
                 [/not]
             [/filter]
             {UPDATE_STATS}
+            #Make the event undoable
+            {VARIABLE_OP disallow_undo rand(0..1)}
+            {CLEAR_VARIABLE disallow_undo}
```
Discontinuum, line 534: Why is this needed? I don't see how it's related to recalls

Discontinuum:
I wonder btw if it should be applied to all $allied_sides instead of just side=1

utils/global_events.cfg
Toranks: unit_placed events are always fired on recalls, and before the recall event itself.

Both events do the same thing, [update_stats] if the unit is new and is recalled.

The recall event in practice only serves to clean up starving variable, but I have kept the code for update_stats just in case the unit placed code fails. I can't think of what that case could be, but better to prevent.

Toranks:
> I wonder btw if this event should be applied to all $allied_sides instead of just side=1

I thought the same as you. Let's see if @Dugy  knows why this is so. Otherwise, it should be only side 1 since only side 1 can equip items

utils/global_events.cfg
Discontinuum: You should be extremely cautious about adding random calculations to "unit placed" events because in multiplayer "unit placed" events are not synchronized: when a player opens the item menu, "unit placed" for him maybe called many times as he equips many items, all other players will get all the updates in a single pack and run "unit placed" only once. Are you sure this is absolutely necessary here and won't lead to troubles in MP?

Discontinuum:
> Otherwise, it should be only side 1 since only side 1 can equip items

No, all $allied_sides can. This variable was introduced for multiplayer primarily

Toranks:
I have some lua based code that detects fine whether the player is human or not, no desync issues, I could use it. 

But then I don't understand why you say it should apply only to side 1.

This is the code, by the way:

```
    [lua]
        code = <<
			local result = wesnoth.sync.evaluate_single(
                function()
                    return { value = "no" }
                end,
                function()
                    -- Called only on the client handling the current side, if it is an AI.
                    return { value = "yes" }
                end)
            wml.variables["aww_is_ai"] = result.value
			>>
    [/lua]
```

if aww_is_ai variable is yes, it is ai, if not, it is human

Toranks:
Ah, I see that in multiplayer there are always 5 players, apparently. We wouldn't need to use that check, and it should work fine as-is.

```
    #ifndef MULTIPLAYER
        {DROPS_EXT {CHANCE} {CHANCE_GEM} {WEAPONS} {BOSSES} {ENEMIES} "1"}
    #else
        {DROPS_EXT {CHANCE} {CHANCE_GEM} {WEAPONS} {BOSSES} {ENEMIES} "1,2,3,4,5"}
    #endif
```

Discontinuum:
> Ah, I see that in multiplayer there are always 5 players

In LotI yes, but not in other theird-party dependency addons

Dugy:
Is it really necessary to do this in unit placed? It's a rather error prone thing, because it would call another unit placed in case it modifies a unit and places it with `[unstore_unit]`.

> I thought the same as you. Let's see if @Dugy knows why this is so. Otherwise, it should be only side 1 since only side 1 can equip items

This used to be the concept, but then I changed it to side 1 in single player and sides 1-5 in multiplayer, and that one was changed later because there is a MP mod that adds LotI functionality to any multiplayer scenario.

Toranks:
> Is it really necessary to do this in unit placed? It's a rather error prone thing, because it would call another unit placed in case it modifies a unit and places it with `[unstore_unit]`.

That unit placed event is directly responsible of the desynchronization, because it is always executed before the recall event. 
Avoiding undo will never cause a double event error, don't worry about it. Also, both events filter for the variable `updated=yes`, and only one of the two will be executed because the first one will change that variable to yes.

Dugy:
That looks like a fine trick.

Toranks:
I have tested this semi-hidden lua function on the advice of a developer, and it works on my addons, on 1.16 1.17 and I just tested it on LOTI and it works too.

Dugy:
True. But I merged it already, so let it be there. I have recommitted that one commit you accidentally reverted.

PR #562, opened 2023-07-09
Fix broken tags in REMOVE_TRAIT_BY_ID

Discontinuum:
Hm, it happened still not fixed. Also I moved it to utils.cfg, I think it makes there more sense than in abilities.cfg


Dugy:
Oops, sorry, I forgot what I read and merged it.

PR #564, opened 2023-07-09
Nightmare Cellar

white-haired-uncle:
Created RARE_ITEM_TURN allowing setting of turn value for dropped item

I don't think I like optional arguments passed as an ordered list (to lua function).  If I were doing this from scratch, I think I'd rather pass a table with arguments declared explicitly.  But I don't have the XP to make that argument now, best to leave it alone.

Ref #527

### DIFF
--- a/lua/items.lua
+++ b/lua/items.lua
@@ -378,7 +378,8 @@ end
 
 -- Place item on the ground at coordinates (x,y).
 -- Optional parameter crafted_sort: if present, overrides item_sort of the item.
-loti.item.on_the_ground.add = function(item_number, x, y, crafted_sort)
+-- Optional parameter turn: if present, overrides current turn.
+loti.item.on_the_ground.add = function(item_number, x, y, crafted_sort, turn)
 	local record = {
 		type = item_number,
 		x = x,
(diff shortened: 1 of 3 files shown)


Discontinuum:
It accidentally reverted changed from #559

Dugy:
Optional arguments for backwards compatibility can be a bad direction, but it's fine if it's just one.

white-haired-uncle:
> It accidentally reverted changes from #559

I saw that, and I thought [6adba4d](https://github.com/Dugy/Legend_of_the_Invincibles/pull/564/commits/6adba4d78e93770fe4fdccbb3963fcb78271729e) would fix it.

Sorry, still learning.

white-haired-uncle:
> Optional arguments for backwards compatibility can be a bad direction, but it's fine if it's just one.

There was one opt arg already.  I added a second.  That I don't like.

Discontinuum:
Do we need the same kind of protection for Chapter 5 armouries? For other Inferno loot rooms?

Dugy:
I think it might be useful.

white-haired-uncle:
Looks like it for 3 storerooms in ch5 at least.  Working on it.

Surprised I never noticed this before.  I leave the room on turn 1 very often (switch rooms back and forth to move items around units in my autorecall list).

PR #566, opened 2023-07-09
add RARE_ITEMS_TURN to ch5 storerooms

white-haired-uncle:


Dugy:
I'd put -9000 there just because of the meme, but fine.

PR #567, opened 2023-07-09
Fix experience by devastating blow

Toranks:
I thought "no" was the default option, and I have to explicitly declare it. I have edited the wiki to reflect this circumstance.
Related with #525 
https://wiki.wesnoth.org/index.php?title=DirectActionsWML&type=revision&diff=71498&oldid=71342


Dugy:
Good find.

PR #569, opened 2023-07-10
fix steelhive portals

white-haired-uncle:
In Steelhive, only one portal was active.

All of the portal events had the same id.  

I included the side in the event id, which may be redundant, but I suppose multiple sides could have a portal at the same x,y?

Also, some portal/tombstones were spawning enemies in walls, where you could not hit them.  I only updated code to use this where I could test at the moment, I suspect it might be used elsewhere as well.


Dugy:
How were they spawning these enemies to walls? Were these portals inside walls or it happened only if the hex was occupied?

white-haired-uncle:
The portal location was fine, but if there were units around the portal they'd spawn in the walls, especially if the portal was in a narrow corridor.

Dugy:
Okay, thank you.

PR #570, opened 2023-07-10
Logical encampment building

Toranks:
At this time, the camp can be built anywhere on the map, even in fog and darkness on the other side of the map.
This is very prone to exploits, so I came up with a simple and logical way, which requires no change in the feature description.
You can simply build the camp if there are any friendly units but NO enemies inside the camp to be built.

https://github.com/Dugy/Legend_of_the_Invincibles/assets/40789364/2148a718-e850-4572-914e-c0ae6b5e8261


Dugy:
Yeah, sounds reasonable.

Issue #435, opened 2022-11-21
Side story campaigns do not have added unit advancement paths


matsjoyce:
LotI has a macro `LOTI_EXTRA_ADVANCEMENT_LINES` in it's _main.cfg, while The Beautiful Child and Affably Evil do not. This results in normal units like Elvish Sharpshooter going into the normal AMLA cycle, not progressing into the LotI added unit types. Is this intentional? (I would submit a PR, but those campaigns aren't on GitHub IIRC)

Dugy:
Okay, so I created a `spinoffs` subfolder and put these campaigns there (I set the folder to be ignored by the server). Then I fixed this bug.

PR #571, opened 2023-07-10
Typo on the Beautiful Child textdomain

Toranks:
Using the improved wmlgettext tool from the latest version 1.17.18, I have found several typos.

By the way, are these three separate campaigns still being updated or are they already obsolete?

### DIFF
--- a/spinoffs/The_Beautiful_Child/units/Magic_Missile.cfg
+++ b/spinoffs/The_Beautiful_Child/units/Magic_Missile.cfg
@@ -1,4 +1,4 @@
-#textdomain wesnoth_tbc
+#textdomain wesnoth-tbc
 [unit_type]
     id=Magic Missile
     name= _ "Missile"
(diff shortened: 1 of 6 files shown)


Dugy:
They are not obsolete, but they generally get less attention.

Discontinuum:
TBC on the server right now is of an older broken version without LotI units advancements recently was noted on the forum https://forums.wesnoth.org/viewtopic.php?p=682529#p682529. Maybe KtK or AE have the same issue

In repo it was fixed in #435 but you didn't upload the fix to the addon server

Dugy:
I could upload them, but would they work with the server version of LotI? Any idea if it's necessary to update LotI as well? I feel a bit uncertain about uploading it now when the Spanish translation by @Toranks is half-done.

Discontinuum:
Personally I wouldn't mind to update LotI itself, there is a number of patches accumulated, but I would like probably to wait a bit, like a week or so, to make some more patches and to prepare my mods to the update

Toranks:
The extra campaigns have another textdomain and I'm only translating wesnoth-loti. It does not matter if you update them before or after, since I am not translating the story and dialogues.

Dugy:
> Personally I wouldn't mind to update LotI itself, there is a number of patches accumulated, but I would like probably to wait a bit, like a week or so, to make some more patches and to prepare my mods to the update

All right. Perhaps I'll force myself to add some code here myself like in the times of old.

PR #573, opened 2023-07-11
Suck and spell suck specials table insert

Toranks:
Same method as #530 

Suck is a stackable ability, which can have different values on different weapons, or affect only some weapons. Until now the only way to see it was on the unit info page, but it's more comfortable that way.

imagen
imagen

### DIFF
--- a/lua/stats.lua
+++ b/lua/stats.lua
@@ -258,7 +258,7 @@ function wesnoth.update_stats(original)
 		if weapon_type == "magic" then
 			weap.damage = weap.damage * magic / 100
 			if spell_suck > 0 then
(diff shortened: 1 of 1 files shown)


Toranks:
By the way, @Dugy suck works only on impact, pierce, blade, and arcane according with the ability event. So if you change the element of the weapon with a second item (for example to lightning or ice), it no longer works. Is that intentional? Should it be fixed?

Dugy:
> So if you change the element of the weapon with a second item (for example to lightning or ice), it no longer works.

Yes, it is intentional. There is even an item that changes the damage type and adds suck, and you have to change damage type if you want to use the (relatively powerful) item to full extent.

PR #574, opened 2023-07-11
remove FOREACH in chapter 9

white-haired-uncle:
Removed FOREACH macros

Added MOVE_SIDE_TO macro for code which was used a few times, particularly when Romero shows up.

I think there might be an unrelated issue in chapter9_utils, STATUE_AMBUSH NUMBER.  One or two variables (gender_...) get set in prestart, then cleared in moveto.  What if you don't "moveto"?  No big deal, just wondered why they were hanging around in debug>inspect (also wonder if the seal guardian stores need to be kept once they are defeated).


Dugy:
I have no idea what did I mean with that moveto event.

Issue #560, opened 2023-07-09
Subtle problem caused by the use of get_type_advancement in effect_containers (unitdata.lua)


Discontinuum:
So, https://github.com/Dugy/Legend_of_the_Invincibles/blob/2e08f7f943fef05ed2260b198ee0155a8aa5a2a5/lua/unitdata.lua#L60-L77 ``get_type_advancement`` local function gets the contents of an advancement from the unit model (AdvancingType or Type) by id, if there is no such advancement, it returns the table with just its id. 

I don't understand though why it's used here in effect_containers https://github.com/Dugy/Legend_of_the_Invincibles/blob/2e08f7f943fef05ed2260b198ee0155a8aa5a2a5/lua/unitdata.lua#L240-L247

We have already got the contents on line 240, why would we need to override it on 246? It would only make sense if actual AMLAs of units were stored with just their ids in "modifications" section, so we had to go in the model to parse the effects. But actual acquired AMLAs always stored in their full form (otherwise Wesnoth wouldn't see default effects like +hp, +damage, etc) if I understand correctly.

Where this causes trouble: right now I'm developing a mod that allows all units to benefit from LotI AMLAs and books (because right now if a unit doesn't have a special LotI upper level, neither legacies nor books can be used. Like with the Steel Scourge that you free in the prison in Akula's dungeon). My idea was to consider which legacy and which books this unit has and dynamically add required advancements in a "pre advance" event. The problem is that because of the use of ``get_type_advancement`` many interesting LotI effects, like penetration or bonus attacks, don't work: ``get_type_advancement`` can't obviously find the newly added advancement in the non-LotI unit type, it returns ``{ id = advancement_id } `` table and stats.lua don't see effects. If I hack the save file and turn all those added advancements into objects , everything works as it should (because for ``[object]`` we only mess with ``contents`` only if it's an item that can have set effects).

Discontinuum:
Hm, I see that this model was first introduced in this commit for "effects" https://github.com/Dugy/Legend_of_the_Invincibles/commit/863a49884b5e81fcca5583693b96d075f209d39c (and later "effect_containers" was coded by analogy). The description exactly says that "the idea is to remove full [advancement] tags from the units at some point in the future". But 4 years passed and this future didn't come. Should we still stick to it?

Dugy:
[...]

PR #576, opened 2023-07-12
Only request data about advancement if it's in the bare form 

Discontinuum:
@edwardspec  and @Dugy that is my touch on the problem that I can propose to your consideration. 

Logic is pretty simple: we check that advancement in the form ``{id = adv_id}``, if it is, we use ``get_type_advancement`` to try to get the full info. If it's not, then it must be the fully spelt advancement, so we just use ``contents`` without touching it. 

I also want to show you the mod (merely a draft but still) that I'm implementing that I mentioned in #560 and which stumbled upon this bug. https://github.com/Discontinuum/LotI_Legacies_and_Books

The mod adds advancements to units that don't have LotI version. It's planned to add all needed legacies and books, for now only fire and ice dragon legacies are implements. 

``events.cfg`` sets up variables containing WML code of advancements (initially I used events to add amlas, so the name is a bit outdated, will rename to amlas.cfg or something like that). As I said in the comments under #560, I gave up on the casting unit's animations and just deleted all the code that mentions it when copypasted it from LotI.

The lua code that runs during ``pre advance`` is pretty straight-forward: we create a set of the unit's present advancements and check if it has a legacy. If it has a legacy, we add advancements of that legacy from the corresponding variable. The same logic will be applied to books: it will check which books the unit read and will add corresponding advancements in the same way.

### DIFF
--- a/lua/unitdata.lua
+++ b/lua/unitdata.lua
@@ -55,6 +55,17 @@ local function wml_modification_iterator(unit, tag, filter)
 	end
 end
 
+-- Check if the advancement's table has only the id key {id = adv_id} with no extra info
+
+local function is_bare_advancement(adv)
+	for k in pairs(adv) do
+		if k ~= "id" then
+			return false
+		end
+	end
+	return true
+end
+
 -- Helper function to obtain a unit type's advancement with a given id
 
 local function get_type_advancement(unit_type, advancement_id)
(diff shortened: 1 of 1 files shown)


Dugy:
If I accept it, your add-on will break as soon as someone finishes the transition to using thin advancements in units.

Discontinuum:
> your add-on will break as soon as  someone finishes the transition to using thin advancements in units

Hm, would it necessary mean the full advancements not supported when the transition is finished? I don't understand why the future thin advancements implementation couldn't use the check like the one in this patch

Dugy:
Well, keeping support for that should not cost much.

Dugy:
Please fix the luacheck problem and I'll merge it.

lua/unitdata.lua
```diff
@@ -242,7 +253,7 @@ local wml_based_implementation = {
 			if modif_type == "object" and contents.number then
 				-- This is an item, therefore we must add "item set" effects (if any).
 				contents = loti.unit.item_with_set_effects(contents.number, set_items, contents.sort)
-			elseif modif_type == "advancement" then
+			elseif modif_type == "advancement"and is_bare_advancement(contents) then
 				contents = get_type_advancement(unit.type, contents.id)
```
edwardspec, line 257: space before "and"

PR #579, opened 2023-07-12
Chapter10

white-haired-uncle:
The ch9 commit doesn't belong.  Figured out most of what I was doing wrong, still not quite there yet I guess.

Tweaked the first few scenarios to make them at least playable.  Probably still too hard, but better to aim small.

Some of the later scenarios are still ridiculous, but I hate ch10 too much to test any changes beyond just making sure the for/each loops work right.

scenarios10/01_Despair.cfg
        - lower demon strength (better, but still pretty hard compared campaign to date)
        - comment about removing set_menu_item "hacks"
        - moved a demon who ended up in the wall when V cast her spell at the beginning
        - replace FOREACH

scenarios10/02_And_Another_Orcish_Assault.cfg
        - lower demon strength (better, but still way too hard)

scenarios10/03_Fimbulwinter.cfg
        - lower demon strength (better (winnable), but still quite hard)
        - replace FOREACH

### DIFF
--- a/scenarios10/01_Despair.cfg
+++ b/scenarios10/01_Despair.cfg
@@ -77,6 +77,7 @@
             image="wesnoth-icon.png"
             message= _ "Meanwhile..."
         [/message]
+        # this shouldn't be here anyway, but why not just clear_menu_item to be sure?
         [set_menu_item]
             id=5confautorecall
             [show_if]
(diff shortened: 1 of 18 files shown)


Dugy:
Thanks.

I am not sure why was I doing that set_menu_item hack, maybe remove_menu_item wasn't there when I was writing that part.

Toranks:
[clear_menu_item] In any case, the method you use works the same, except that [clear_menu_item] allows you to clear several menus in a single command separated by commas.

PR #580, opened 2023-07-12
Fix animations on corrupted elven scout and rider

Toranks:
Move, defense, and melee attack of Corrupted Elven Scout and Elven Rider are broken. This fixes it adapting from the originals. Maybe it uses old animations and you forgot to update.

imagen


Dugy:
Yeah, this happens when unit images are moved between versions.

PR #581, opened 2023-07-12
All these years nobody knew that all those boots (and a couple of gauntlets) were supposed to have different colours

Discontinuum:
I accidentally noticed ``error engine: Invalid color in palette: magenta-`` in the log, and it turned out that all those items, mostly boots were supposed to have different colours, but it was spelt "magenta->colour" instead of "magenta>color' so they were all magenta


Dugy:
Good eye.

PR #582, opened 2023-07-13
Cover case when Harry dies before reach the town

Toranks:
In discord someone has reported Harry's death off screen (in the shroud) and when they get to town they talk to someone who doesn't answer. With this change, Harry will still speak as first option (tested), but in case of his absence, any of the surviving soldiers from side 2 will do it, and the conversation still makes sense as is.

I think it's almost impossible for them all to die, so this simple change will suffice.

### DIFF
--- a/scenarios2/02_Gods.cfg
+++ b/scenarios2/02_Gods.cfg
@@ -481,7 +481,7 @@
             message= _ "This must be the town of Aloh."
         [/message]
         [message]
-            speaker=Harry
+            side=2
             message= _ "The elves are coming back! To your posts, men!"
         [/message]
         [message]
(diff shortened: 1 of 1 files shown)


Dugy:
Won't that possibly pick a random side 2 unit instead of their leader even when the leader is alive?

Also, if you want to be absolutely sure there would be a side 2 unit, you can spawn one at the start of the event.

Toranks:
< Won't that possibly pick a random side 2 unit instead of their leader even when the leader is alive? >

No, the leader is always chosen first if he is present. They are chosen in order of creation, not random. And I already verified that it worked.

Dugy:
All right. I will merge it whenever you want.

Toranks:
To make sure, I did what you said. The soldier spawns in the keep, and I've placed myself right there with Efraim to make sure he doesn't overwrite a possible friendly or enemy unit there. 

It's very common for Harry to die, it seems. There are several reports on #discord. Unlikely to all die, but it's good to make sure.

imagen

Dugy:
All right.

PR #583, opened 2023-07-13
Chapter6

white-haired-uncle:
Ch6, S1 is absolutely brutal.  I have tried tons of approaches, and found no way to get through w/o a load of save-loading.  Issues include:

        - must know (guess) location of portal
        - must know (guess) right amlas
        - must never get unlucky
        - probably need a little luck of your own (good item drop)
        - your level is too high (giving too much XP to enemies)
        - blades vs undead

Move the Revenant back a turn, so you don't get ambushed coming out of the portal, and you have at least a chance of making it to the coridoor before being surrounded.

HARD
        - drop one heal potion
NORMAL
        - drop two heal potions
EASY
        - make them greater healing potions

With these changes, I was barely able to get through on normal.  On turn 4, I lucked into a Silver Axe, which I thought would break my test (a little luck can make a huge difference).  But then on about turn 6, three dead hit me with all their attacks while did absolutely nothing in retaliation, and I ended up having to use the heal potion.

I don't think these changes are enough, but a step in the right direction.

Scenario 2 is also very hard if you haven't played it before, and very hard even if you have.  I didn't touch it, hopefully you carry over a heal potion from S1 and that will help.

scenarios6/14_Frozen_Land: Replaced FOREACH.  I don't have a save to test this.


Toranks:
Managing retaliation weapons is important in this part of the campaign as well. I suggest that the maximum difficulty stay as it is. It's very rare that the maximum difficulty can be cleared without restarting once, but it seems good to me.

Discontinuum:
Hm, strange, Ch6S1 never seemed to me particularly hard. You may get low on turns, but other than that, I dunno. It may be challenging, but brutal, I wouldn't say, it's not ch10

Discontinuum:
> I suggest that the maximum difficulty stay as it is

I guess I second this, I remember dalong was planning to make an even more hard Nightmare difficulty level, there are some people for whom even the current Hard is too easy

white-haired-uncle:
P.S.  Nightmare mode - I think a player could activate this for themselves by either refusing to use items, or perhaps only using crafted items.  Maybe never using redeem.  Etc.  I'd certainly give N-mode a try if someone eles made one, but I'd try other things before going to all that work if I was thinking about coding it.

Discontinuum:
[LotI2-The Awakening replay 20230713-185430.gz](https://github.com/Dugy/Legend_of_the_Invincibles/files/12040737/LotI2-The.Awakening.replay.20230713-185430.gz)

No reload run on the current master, hard, reloading-proof rng. Yeah, I got lucky at times, I wouldn't recover from Wraith if I didn't have a potion (or if I knew about the Wraith, I could've been more careful). I wanted to go to the loot room in the bottom left corner but lost time in another empty chamber, still it turned out okay on turns

Discontinuum:
For the first timer you most likely need reloads, I agree. But from my memories, when I was doing LotI first time, I didn't note this scenario as particularly hard. I struggled very badly in the 7th chapter, but not here

Discontinuum:
Hm, I looked at the map without shroud, it turns out there is a whole loot room in the right bottom corner which I never found. 

About Wraith and possible wrong directions, maybe we can add some dialogues to hint/warn on that

Discontinuum:
I've watched your replay from here https://forums.wesnoth.org/viewtopic.php?p=669063#p669063 and indeed you struggled a lot, getting caught in bottlenecks and near-death a lot and completed on the last turns. I think maybe what we do differently is that I take faerie fire advancements first. You took two times nimble and then only axe advancements, and many times you couldn't finish them with weak faerie fire. I take faerie fire a lot and just walk through them. Lich without faerie fire advancements is also kinda problematic, while developed faerie fire with luck may kill it in one attack. I'm pretty surprised tbh why you don't take a faerie fire grade at least once

> P.S. There's a scenario, I can't remember which, but you start on an unmarked keep 

It probably was https://github.com/Dugy/Legend_of_the_Invincibles/blob/master/scenarios2/09_Underground.cfg

white-haired-uncle:
At the time, I didn't know about the short cut (IIRC) and was always pressed for time, hence nimble.  And I probably took axe because I was taking so much melee damage I want to give some back.

Also, I used to like "harder to hit" early on when E/L are very weak (recently I've decided that's probably a waste, as once you get to later levels pretty much all enemies have magical, which I assume negates harder to hit).  So maybe I was also focusing on taking amlas for more than just making it though this scenario, which I've decided is unfortunately pretty much required.

Lately, I focus on ffire in S1, with one movement thrown in so that I can approach an enemy and use my weapon of choice (ffire).  I've noticed that often they won't approach me if I am just outside their attack range, so I end up with no choice but being the one to make the approach, and I need to be the attacker lest they clobber me with melee.

Yes, it was Underground, thanks.  Don't know why the wiki search didn't come up with that one when I search for glyph, it's right there in the walkthrough.

Discontinuum:
I think I usually tried to farm this scenario for as much xp as possible, even if it caused me extra reloads

white-haired-uncle:
I could get behind #587 instead of potions.

I would like to move the revenant though. If L is low on HP and triggers the appearance of E using penetrates, she could get killed by that revenant pretty easily after she is moved to within its striking distance.

Dugy:
Guys, decide on something you'd both like, I usually prefer to solve problems with creative approaches that may or may not work.

Discontinuum:
I can propose the following things:

1. Merge buffs #587
2. Revenant replaced as @white-haired-uncle proposes
3. Add some dialogue lines to hint the player that they are moving away/not the shortest path, but that there might be treasures here. Also add some cautioning dialogue about the wraith in the shroud (near the exit)

Additionally I can think of
4. Add a button 23,18 as shown in the picture that Leth can press to crumble the walls. But place some more loot in the long way. Btw even right now this way works: if Lethalia steps on the hex I marked "button", it triggers the "Here he is!" event (radius of the filter in the event captures that hex)
изображение

5. Maybe add some button here (27,32) as well, to have easier access in the bottom room, but how you would pillage it if you still have to get Efraim up there is a question
изображение

Dugy:
The scenario with runes marking the correct path is Underground in chapter 2. The runes could be used to mark which way is correct, but maybe extending the time could be better than that.

white-haired-uncle:
The one time I was able to pillage I had picked up some really good item(s), holy sword and wanderlust/taxes maybe, something like that, and pretty much finished off the undead early, then sent one unit exploring while the other waited to make a run at the exit.  Probably ran right past the lich to escape.

white-haired-uncle:
New scenario version
- moves revenant
- adds glyphs, 
      - hopefully makes pattern pretty obvious if you're looking for it
      - should draw player into side room with secret passage
      - might defer player from wraith
- secret passage event
      - makes visible passage for direct route to E
      - adds a couple more runes to keep player on track (the 'wrong way' rune should not exist if player doesn't use secret passge)

New map 
- adds ruble as subtle hint to secret passage
- map editor clearly reformats file -- I only added ^Xos to a few tiles to mark passage

Dugy:
Does the rest of the people involved agree on this solution? @Discontinuum?

white-haired-uncle:
(these are in addition to, not replacements for, #587)

Discontinuum:
I'm going to review it a bit later today

Discontinuum:
Looks fine, I don't mind to merge this and #587 , if we merge, I would probably make some minor improvements (dialogue line warning about the wraith in the distance and some animation maybe for crumbling walls)

Discontinuum:
By the way, there is a subtle problem that Efraim can't change weapons/armour if Lethalia had luck to pick up some swords/good armour, maybe we should add a castle tile somewhere after we found Efraim

Dugy:
Yeah, adding a castle tile somewhere sounds reasonable.

Dugy:
I have added some castle tiles myself.

PR #586, opened 2023-07-13
Make potions' flavours visible

Discontinuum:
I remember long time ago they were visible. now only description


Dugy:
Yup, that was a bug.

PR #587, opened 2023-07-13
Buff weaked Efraim and Lethalia

Discontinuum:
Considering the discussion in #583, may propose this, buff them a bit (+5 hp, less xp to level on lower difficulties) and add one strike to basic faerie fire of Lethalia, that should prevent her from being stuck too much

### DIFF
--- a/units/Efraim_lich-weakened.cfg
+++ b/units/Efraim_lich-weakened.cfg
@@ -6,7 +6,7 @@
     gender=male
     image="units/important/efraim.png"
     profile="portraits/Efraim-later.png~RIGHT()"
-    hitpoints=55
+    hitpoints=60
     movement_type=elusivefoot
     [movement_costs]
         forest=1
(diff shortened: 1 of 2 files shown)


white-haired-uncle:
Tried this and liked it (normal diff).  I did get lucky with some decent item drops (of course, always happens when I'm testing).  Was concerned about the extra ffire attack affecting later scenarios, but they reset in S2 I believe.

Toranks:
So this change is only for the first scenario and stats are changed on next scenario? An extra hit for Lethalia is devastating later on.

Dugy:
This changes only the weakened type, that is transformed to the normal type in scenario 2.

Dugy:
Well, I was assuming that this was resolved by white-haired-uncle's pull request that altered the first scenario.

Yes, I could update the server version, there wasn't much stuff going on here, so I would not interrupt anything probably.

white-haired-uncle:
My change did not include this.  This one should be merged, IMO.

Dugy:
Yes, but it made the scenario easier in different ways. Or do you mean that it was also assuming these changes would be made?

Dugy:
All right.

PR #588, opened 2023-07-13
Ally mages with all player enemies except trolls

### LINKED ISSUE #584: Chapter 2 Scenario 2 "Gods" can be defeated offscreen
The trolls have force to hit 0 mages, but any other side could kill Jacob and produce defeat before even reaching the mountains. Since it is not explained anywhere that saving him is necessary, the player does not know why he has been defeated, and the conversation of Jacob's death makes no narrative sense, about something that happens out of sight of the player.

There would be two solutions:

- Force_to_hit other sides against mages

- Have the AI avoid entering the mountain, at least until the player meets Jacob.

On the other hand, this code doesn't seem to work as intended. What is the original intention to this AI aspect? All enemy AI sides have it:

```
            {MODIFY_AI_ADD_ASPECT 4 attacks (
                [facet]
                    name=testing_ai_default::aspect_attacks
                    id=dont_attack_others
                    invalidate_on_gamestate_change=yes
                    [filter_enemy]
                        [not]
                            side=2
                        [/not]
                        [not]
                            side=3
                        [/not]
                        [not]
                            side=5
                        [/not]
                        [not]
                            side=6
                        [/not]
                        [not]
                            side=7
                        [/not]
                        [not]
                            side=8
                        [/not]
                    [/filter_enemy]
                [/facet]
            )}
```

Toranks:
Resolves #584 

The alliance is hidden, not visible to the player, so it's the easiest way without making it look like the mages are ACTUALLY allied.

I have also fixed the name of the AI aspect, which is old.
What this AI does is that "Specifying scouts not attack whilst scouting" according to the wiki:
https://wiki.wesnoth.org/Formula_AI_Code_Library#Specifying_scouts_not_attack_whilst_scouting
(Keep in mind that it is more than 10 years old)
But it does not prevent them from attacking each other. That's ok. There will be a fight, since they are really enemies between them, but it will be minimized, since they will prefer to explore except to attack side 1.

### DIFF
--- a/scenarios2/02_Gods.cfg
+++ b/scenarios2/02_Gods.cfg
@@ -58,7 +58,7 @@
         random_traits=yes
         side=3
         canrecruit=no
-        team_name=good
+        team_name=good,Saurians,Undead,Bandits
         user_team_name=_"Good"
         gold=0
         income=−2
@@ -96,7 +96,7 @@
             leader_value=100
             {MODIFY_AI_ADD_ASPECT 4 attacks (
                 [facet]
-                    name=testing_ai_default::aspect_attacks
+                    name=ai_default_rca::aspect_attacks
                     id=dont_attack_others
                     invalidate_on_gamestate_change=yes
                     [filter_enemy]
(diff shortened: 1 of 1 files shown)


Dugy:
I think you may now safely delete all the MODIFY_AI_ADD_ASPECT blocks.

Toranks:
Ok, also some were inside an [ai] tag and some were not. They are supposed to go inside [side] [ai] according to the documentation, so some of them have never worked xD

In any case, it is no longer essential, so it is better to delete it.

Dugy:
> Ok, also some were inside an [ai] tag and some were not. 

Ooh, that explains a lot.

Issue #577, opened 2023-07-12
Consider switching to wesnoth.effects for some properties


Discontinuum:
Some properties happen to have a built-in support in wesnoth.effects from 1.14, so maybe we can switch to using built-in ones (although the code in stas.lua probably have to stay for supporting legacy saves).

Among them are vision (``apply_to=vision``), max_attacks (in docs ``max_attacks`` with ``increase=`` key while LotI uses ``add``), alignment (LotI items use 
```
[effect]
apply_to=alignment
alignment=chaotic
[/effect]
```
while Wesnoth docs define the built-in alignment effect as
```
[effect]
apply_to=alignment
set=chaotic
[/effect]
```

Dugy:
Yes, that would make sense.

PR #589, opened 2023-07-15
Use built-in wesnoth [effect] for alignment and max_attacks

Discontinuum:
This is the first patch about #577 making alignment and max_attacks using built-in wesnoth.effect implementation. Vision stays the old way for now, I need to test more how it will work in older replays


Dugy:
This simplifies things a little bit.

PR #592, opened 2023-07-16
remove FOREACH, use [is_in_list]

white-haired-uncle:
Replace FOREACH statements

use new [is_in_list] from lua/utils.lua I improperly added yesterday to collapse big if/or block.

Tested over several advancements in part 1.


Dugy:
Luacheck is complaining about mixed spaces and tabs, please fix that and I will accept this.

PR #599, opened 2023-07-17
add [sort_unit_list]

white-haired-uncle:
A few times in ch4 you can recall previous units.  This sorts the list by name.

Will use this to clean up autorecall a bit once I get to ch5 and can test.

### DIFF
--- a/lua/utils.lua
+++ b/lua/utils.lua
@@ -68,3 +68,30 @@ function wesnoth.wml_actions.is_in_list(cfg)
 	end
 	wml.variables[to_variable]=ret
 end
+
+-- [sort_unit_list] Sorts array of units by name/type
+-- Requires unit_list=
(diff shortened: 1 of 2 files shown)


white-haired-uncle:
Hmm, sorting Valhalla affects Soulgate.

I don't see this as a problem, as it's kind of random (until players figure it out and start renaming units to game the system).  But if it is, I can use a sorted list for the selection process, leaving the Valhalla order untouched.  I saw a discussion recently about which order units should be pulled with Soulgate...

https://forums.wesnoth.org/viewtopic.php?p=682063&hilit=Valhalla#p682063

@Toranks

Toranks:
"Not that it really matters. Soulgate doesn't have a "cooldown", so you can just go back and forth between scenarios one turn 1 casting sg and running away."

I noticed this when I was already on the antediluvian tunnels of Akula, but the influence is greater in the earlier levels.

Issue #601, opened 2023-07-17
Akula's Place of infinite gold


white-haired-uncle:
Every time I enter Akula's Place, I get +200 gold.  Every time.  Over.  And.  Over.

Seems to me, just drop that -- you go with the gold you brought with you.  Or maybe you get 200 ONCE (but no, this scenario is already WAY too easy, even if you don't bother with the amps and such and just go straight at her with Dop's and potions).

white-haired-uncle:
Cool, but I can still beat it on turn 1.

4 speed potions, a couple greater healing, and E's wave of fear.

Of course, she only has 180HP, though it looks like you were trying to increase her from 500 to as much as 1000?  Might need a madman's alacrity or two in that case.

Dugy:
With the four amplificators destroyed, she's not supposed to be very strong (but it looks like she becomes a joke of a boss). Also, it looks like you're playing on Easy, and I made her stronger only on Normal and Hard. Unless I screwed up somewhere.

white-haired-uncle:
I'm playing on normal, I didn't test your change and I missed the fact that destroying amps (severely) knocks down her HP (I thought it just reduced that lightning damage that comes out of nowhere).

E starts with fear 15x7 and 10 cold pen from Diadem of Dark.  Four speed potions and a creepy potion later, 15x11 with 30 cold pen.  Then, just to see what I could do, I craft misanthropia sword/knife, giving me fear 21x11 and 70 cold pen.  One greater healing to recover my movement, and she takes 192 HP in a single attack on turn 1.  Even if she had a lot more HP, I've still got another 5 greater healing potions.

Yeah, this is going a little all out.  Normally, I don't mess with potions, and still win around turn 6 or 7 using just E/L and their 20 or so dopplegangers (which, IMO, shouldn't be allowed on the autorecall list and should cost like 100 to recall, if you're even allowed to recall them at all -- they can be ridiculously valuable if you gear up and especially take the right potions, which don't wear off when they advance, just before they are created).

I'm thinking maybe make the "moat" around her island impassable even by flyers and remove the bridges.  Each of the reprobates you kill opens up a bridge to the island.  And the reprobates are surrounded by enemy units from the start so you can't just exploit greater healing potions to kill one on turn one and open a clear path to Akula.

Dugy:
I think a much easier option would be to have Akula start with some units already on all castle tiles.

The doppelgängers were not intended to ever get to the recall list. That is a bug.

[...]

PR #604, opened 2023-07-18
remove dopp carryover

white-haired-uncle:
Tested with army is born / dealing with trolls.  They're not there in invasion.

This is kinda big.  And good.

Ref #601

### DIFF
--- a/utils/global_events.cfg
+++ b/utils/global_events.cfg
@@ -421,6 +421,8 @@
 
     {LOTI_MAKE_DOPPELGANGER}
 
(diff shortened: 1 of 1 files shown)


Discontinuum:
Don't you think guys that it was supposed to use older doppelgangers? I always thought it was

Dugy:
I didn't realise this doppelgänger snapshotting could be exploited to the depth of having a strategic depth by itself. Maybe I could once add a way to get one permanent doppelganger or something.

PR #605, opened 2023-07-18
populate akulas place, fix manta id

white-haired-uncle:
Populated castles of Akula and her minions.
Docked their gold to pay for it.
Fixed a couple places where Manta.id was referred to as Manta, not akulas_sister.  She now autorecalls automatically, as I believe was intended.

I could lower the effect of destroying amps on Akula's HP from 40% to 20% while I'm in there.

Ref #601

### DIFF
--- a/scenarios5/23_Akulas_Place.cfg
+++ b/scenarios5/23_Akulas_Place.cfg
@@ -82,7 +82,7 @@
         side=2
         canrecruit=yes
         recruit=Javelineer_steel,Dark Sorcerer_steel,Duelist_steel,Longbowman_steel,Swordsman_steel,Lieutenant_steel
-        {GOLD 400 450 500}
+        {GOLD 200 250 300}
         {INCOME 30 40 50}
         team_name="Dreadful Evil"
         user_team_name=_"Dreadful Evil"
@@ -96,7 +96,7 @@
         side=3
         canrecruit=yes
         recruit=Spearman_steel,Dark Adept_steel,Fencer_steel,Bowman_steel
-        {GOLD 200 250 300}
+        {GOLD 100 150 200}
         {INCOME 20 30 40}
         random_traits=yes
         team_name="Dreadful Evil"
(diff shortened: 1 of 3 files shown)


scenarios5/23_Akulas_Place.cfg
```diff
@@ -230,6 +230,54 @@
         {VARIABLE last_scenario 23}
         {VARIABLE amplificators_destroyed 0}
         {MODIFY_UNIT side=7 status.petrified yes}
+        {VARIABLE_OP demon_type rand (Javelineer_steel,Dark Sorcerer_steel,Duelist_steel,Longbowman_steel,Swordsman_steel,Lieutenant_steel)}
```
Dugy, line 233: Why `demon_type`? They are not demons.

scenarios5/23_Akulas_Place.cfg
```diff
@@ -230,6 +230,54 @@
...
+        {GENERIC_UNIT 2 $demon_type 23 14}
+        {GENERIC_UNIT 2 $demon_type 24 16}
+        {GENERIC_UNIT 2 $demon_type 26 14}
```
Dugy, line 236: This would be far easier with a foreach on the result of `[store_locations]` with a filter for castle terrain in some radius around Akula.

white-haired-uncle: I'll check that out, thanks.

white-haired-uncle: Um, easier?  No.  Not.  At.  All.

If the syntax for [store_locations] wasn't insane enough, the gui doesn't distinguish between a lowercase L and a capital i in the terrain type.  Makes a difference though.

BTW, seems strange that Akula's in a lit castle when she's neutral and the player has to actively work at not having a predominately lawful army (Legacy of Light is nice, but with it being just about everyone's legacy along with Prophets and Celestial Messengers being lawful/illuminating...).

Dugy: > Um, easier? No. Not. At. All.

As I know myself, there would be several hexes set wrong and some adjustments to map would break it.

> BTW, seems strange that Akula's in a lit castle when she's neutral and the player has to actively work at not having a predominately lawful army

I think I made it for the purpose of making the player's predominantly lawful army less of a burden in the final fight.

Dugy:
Good job.

PR #607, opened 2023-07-18
new tag [make_units_wander]

white-haired-uncle:
While working on FOREACH, I found several scenarios that "Make guards/demons wander", so I made a function to handle that.

Broke out is_in_list so it could be used for lua as well as wml
Added support for ^ in is_in_list to handle terrain types

Around these blocks there is usually code which looks like it is supposed to make only units that can't see our team wander, mostly by changing the units' side.  But it's done differently just about each time, or at least it looks that way to me, so I didn't want to tackle that without really understanding what was going on.  It might be worth a look, though I played all of these and they play okay.

BTW, while I was testing 10/01_Despair, I saw these dropped at the beginning:
        - 4 scythe, all Exerminators
        - 2 armor, both the same
        - 2 mace, both fighting bull
        - no other item sort had >1 drop

Almost as if the item.sort is picked randomly, and then the item type within that group is always (the first???).  Pretty sure this is new.  I don't think I've touched drops (yet), except adding the fake turn number for Nightmare Cellar.  Could be dumb luck, if I believed in that sort of thing.

### DIFF
--- a/lua/utils.lua
+++ b/lua/utils.lua
@@ -51,21 +51,31 @@ function wesnoth.wml_actions.deny_undo()
     wesnoth.allow_undo(false)
 end
 
--- [is_in_list] tag checks to see if required string= is a member of a comma separated list=
--- Returns true/false (yes/no in wml) in required to_variable.
--- Does not handle empty entries ("" is NOT considered in "cat,,dog")
-function wesnoth.wml_actions.is_in_list(cfg)
-	local list=cfg.list or wml.error("[is_in_list]: missing required list=")
-	local string=tostring(cfg.string) or wml.error("[is_in_list]: missing required string=")
-	local to_variable=cfg.to_variable or wml.error("[is_in_list]: missing required to_variable=")
+-- is_in_list checks to see if required cfg.string is a member of a comma separated l cfg.list
+-- Does not handle empty entries ("" is NOT considered to be in "cat,,dog")
+local function is_in_list(cfg)
+	local list=cfg.list or wml.error("is_in_list: missing required cfg.list")
+	local string=tostring(cfg.string) or wml.error("is_in_list: missing required cfg.string")
 
-	local ret=false
-	for w in string.gmatch(list, "[%w%s]+") do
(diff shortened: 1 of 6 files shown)


Dugy:
Each of these scenarios was implemented at a different time and it's possible I didn't paste the old code but wrote it anew, trying to do it better or something. What worries me more is that there's a large number of changes in what looks like totally unrelated code. What were you doing? Are you sure everything works as it used to?

> BTW, while I was testing 10/01_Despair, I saw these dropped at the beginning:

It comes from a special drop pool (marked as `rare_drop=`) that is reserved for items that are powerful and otherwise rare. They were meant to be treasures placed in hidden caves and such. However, it seems that there isn't enough of items in this drop pool to make duplicates rare, and the Exterminator set is a bit overrepresented there. And it seems that you had a bit of bad luck on top of that. Definitely some of the items at the start of that scenario should be created through the usual item macro, not `{RARE_ITEM}`.

white-haired-uncle:
> Each of these scenarios was implemented at a different time and it's possible I didn't paste the old code but wrote it anew, trying to do it better or something. What worries me more is that there's a large number of changes in what looks like totally unrelated code. What were you doing? Are you sure everything works as it used to?

There's some areas that look like a lot of changes, but are just due to formatting.

The common stuff was to store some units, alter their x,y by random amounts, check that the new locations were within some constraints, and if so move the units.  That's what I replaced.  

I played each of these with some debugging turned on to tell me which unit was considering moving where and if it did so or not.  Any time I saw a unit do something it wasn't supposed to (like going into the woods in Investigations) I was able see why and to fix it.  I also did a spot check on a couple scenarios to see that the visibility/side-change stuff still(?) worked, and it did, but I didn't test all of them because I wasn't entirely sure what each one was trying to do (for example, in one scenario side X appears to move its units on side Y's turn -- I can see visibility check being done there, maybe, but....?).

In the end, I couldn't find any difference between how they work now and how they worked before.

Dugy:
All right. Definitely better than my old code.

PR #609, opened 2023-07-19
fix traders allowing credit

white-haired-uncle:
Buying items from trader was only decrementing the global gold value, not the local gold variable which controlled what you were allowed to buy.

There are too many different trader implementations, IMO.   A project for another day.

### DIFF
--- a/utils/utils.cfg
+++ b/utils/utils.cfg
@@ -195,6 +195,7 @@
             side=$unit.side
             amount=-{COST}
         [/gold]
+        {VARIABLE_OP gold sub {COST}}
     [/command]
 #enddef
 #define SELL_GEMS
@@ -281,6 +282,7 @@ $item_info.description"
                                     side=$side_number
                                     amount=-{PRICE}
                                 [/gold]
+                                {VARIABLE_OP gold sub {PRICE}}
                                 {VARIABLE_OP items_dropped add 1}
                                 {VARIABLE {VARIABLE_NAME} 1}
                                 [fire_event]


Dugy:
I was thinking about creating a lua-based UI for the traders to unify those, but I never got to do it.

white-haired-uncle:
That would be a great first lua gui project.  I have a plan for how I think the autorecall menu should look, but I don't want to do it in WML, and it's a bit ambitious for a first attempt.

Are there any traders where you have them sell particular items?  Like in the arena with Vritra, or the Woodland Cloak in the one with the merfolk vs undead with the skull and crossbones map?  

Are there any traders where certain items have special prices?  Put another way, would it be okay if each item simply had a "trader price" attribute?  I'd expect there would be an across the board price multiplier based on difficutly, but could the base price be globlal?

Dugy:
> That would be a great first lua gui project. I have a plan for how I think the autorecall menu should look, but I don't want to do it in WML, and it's a bit ambitious for a first attempt.

Yes, the GUIs in lua are quite tricky, and it's also easy to create desynch by accident.

> Are there any traders where you have them sell particular items? Like in the arena with Vritra, or the Woodland Cloak in the one with the merfolk vs undead with the skull and crossbones map?

I don't recall putting any items into the traders' stock with some bigger picture in mind except for the Vritra scenarios. The Woodland Cloak may be a coincidence too.

I think the best approach would be to have the trader tag take a list of items that will appear there and a number of random items that would appear with them (their appearance would be saved for the scenario length so that it would not reroll when coming back after getting more gold or something).

> Are there any traders where certain items have special prices? Put another way, would it be okay if each item simply had a "trader price" attribute? 

Random prices can be fine as well, but adding a price attribute to items seems fine too (those without price would not be offered or something). There is already a lua function for generating items from weighted drop pools.

white-haired-uncle:
In the trader menu, gems you can't afford don't show up, but items do.  And if you pick an item that you can't afford, you get a denial message.

Is this what you want, or should they both work the same, and if so which method should be used for both?

I like having everything that is for sale show up, but maybe put items you can't afford in an oblique font.

Dugy:
> In the trader menu, gems you can't afford don't show up, but items do. And if you pick an item that you can't afford, you get a denial message.

Not by design.

> I like having everything that is for sale show up, but maybe put items you can't afford in an oblique font and grey out ok button if selected. 

Yeah, that sounds like a reasonable way how it could be done.

white-haired-uncle:
I'm having a little trouble with item images.  If I include them, the items are spaced too far apart for my taste.  And I'm a bit concerned that if I find a way to scale them down they may not look very good.
Any thoughts on these?

Screenshot from 2023-07-21 06-57-14
Screenshot from 2023-07-21 06-50-31

white-haired-uncle:
Or, if I remove the images from gems only
Screenshot from 2023-07-21 07-13-19

Dugy:
It's possible to crop them, using the `~CROP(` image modification.

Instead of ťhe `OK` button and `Nothing` optional, you can have a `Buy` button and a `Close` button.

Issue #524, opened 2023-06-28
Cleaning up after menu items


white-haired-uncle:
In chapter 8, there are multiple scenarios with a 'Transport units to leader' menu item, all containing (hopefully) the same code.  I decided to try to break this out into a single macro.  I created a macro with a new menu item transport_units_to_Efraim and replaced the [set_menu_item] in 08_The_Desert.cfg with a reference to the new macro.  This worked just fine, except now there were two menu items, 'Transport units to leader' AND 'Transport units to Efraim'.

I finally figured out that since the original menu item was created in an earlier scenario and never removed, it was showing up as soon as teleport_wait was >= 10.  The [set_menu_item]  code I had removed had never even been necessary (perhaps assuming it was the same as in the earlier scenario), as long as there was a teleport_wait variable in use and >=10.  Once I did a [clear_menu_item] on the old one, my new stuff worked fine.

This was kind of hard to figure out.  It seems to me like there are a few options:

1) Do nothing.  Always an option.  It works the way it is (or was in my case).  Redundant, and potentially open to weird behaviour, but working.

2) Leave teleport menu item code in the first scenario it is used, and remove it from subsequent scenarios.  Better, at least code updates only need be made in one place, but still kind of odd IMO (I think I'd feel differently if the scenarios using the menu item were at least contiguous).

3) Add a scenario_end to remove the menu item in each scenario so it doesn't carry over to subsequent scenarios.  If the menu item is going to be defined in subsequent scenarios anyway, this will remove some potential confusion at virtually no cost.

4) Do (3), but using a macro to include the menu item instead of replicating the same code.  Not optimal, because you still need to update event new turn to increment teleport_wait, etc, but IMO the best option yet (save perhaps (1)).

Dugy:
I would prefer one of these two approaches:
1. Put it into a macro, along with the code that removes the menu option at the end of scenario
2. Put it into global events, it will manifest only if the scenario sets the `teleport_wait` variable

Toranks:
Any [set_menu_item] is preserved between scenarios until they are deleted, keep that in mind.

Dugy:
Yeah, but if I recall correctly, setting always the same id will just override it.

white-haired-uncle:
[...]

PR #610, opened 2023-07-19
use macro for teleport_to_leader

white-haired-uncle:
I left in the configurable leader and interval based on difficulty.  Probably won't be used, but it's there, and  {TELEPORT_TO_LEADER_MENU Efraim 10 10 10} provides exactly the same functionality as there was before.

Everything (initialization, teleport_wait++, cleanup, etc) is encapsulated in the macro, save for updating objectives.


Dugy:
Yup, one of the things I should have done ages ago.

white-haired-uncle:
Good exercise for people like me.  I learned a few things about events doing this one. 

Might even help to have a list of little projects like this somewhere that people could do for you.

white-haired-uncle:
Closes #524

PR #611, opened 2023-07-20
rewrite of redeeming event

### LINKED ISSUE #561: Redeem descriptions vs implementation
I had this idea about using lookup tables to get the redeem chance (e.g.chance[redeemlevel][targetlevel]) instead of calculating them.  I don't know that this provides any real benefit besides making some code no one is going to read easier to read.  It would make setting arbitrary rules a lot easier, but is that really going to happen?  Seems like something that would have been a nice idea to begin with, now of limited value.

This lead to a look at the description of redeem (N).  It seems to me there are a few issues related to redeeming units adjacent to the target.

- To match the code, the description for level 5 and 6 should be the same for units adjacent to the target, as should 7 and 8, etc.  Currently, the description for 5 is unique, 6 and 7 match, etc.
- At redeem (15) and higher, there are levels missing from the descriptions
- At redeem (18) and higher, redeeming units adjacent to the target with levels >= 11 are mentioned.  AFAICT, this is never possible.

I think there are a few options
- Do nothing.  I doubt players will even notice anything unless they spend a lot of time in the thing at the end of chapter 9.
- Fix the descriptions to match the code
- Alter the code to match the descriptions (with the exception of the missing levels, no idea what to do there)

white-haired-uncle:
I took the calculations in redeeming and put the results in a lua table of probabilities, then added a couple tags which return yes/no for a redeem on a target.  I think this version is a lot easier to read.

There were some disagreements between the redeem level descriptions and the results of the WML calculations.  I assumed the calculations were always the intended behaviour and produced correct results.  There should be no difference in the results.

The redeem level descriptions have been updated to reflect actual behaviour.

I did a hundred or so redeem attacks with both redeem level <5 and level >5, while watching the lookup results and spot checking the probabilities were correct.  Also advanced a few times.  All went as planned (eventually, took me awhile to understand why you did the check for adjacent units FIRST which seemed kind of backward -- at the time).

Closes #561

### DIFF
--- a/lua/redeem.lua
+++ b/lua/redeem.lua
@@ -565,3 +565,101 @@ function wesnoth.wml_actions.max_redeem_count_from_level(cfg)
     local level = cfg.level or wml.error("[max_redeem_from_level]: missing level= .")
     wml.variables[to_variable] = max_redeem_count_from_level(level)
 end
+
+-- redeem_hit_chance returns the chance of redeem success based on redeem level and target's level
+-- distance indicates whether target is the unit redeemer attacks (distance=0), or a unit adjacent to the
+-- unit redeemer attacks (distance=1)
+local function redeem_hit_chance(redeem_level,target_level,distance)
+	-- nil result indicates no chance
+	local hit_chance = 0
+
+	if (distance < 0) or (distance > 1) then wml.error("does_redeem_hit:  distance must be 0 or 1") end
+	local chance={ 	-- chance to sucessfully hit redeem defender
+			-- index = redeem level, value = table of probablity by target level (from 0)
+		{ 100, 100,  50,  25 },
+		{ 100, 100, 100,  50,  25 },
+		{ 100, 100, 100,  75,  50,  25,  10},
+		{ 100, 100, 100, 100,  75,  50,  25,  10},
+		{ 100, 100, 100, 100,  75,  50,  25,  10}, -- redeem level 5
+		{ 100, 100, 100, 100, 100,  75,  50,  25,  10},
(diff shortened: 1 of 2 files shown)


Dugy:
The values in the probabilities table could be produced by a formula evaluated in a loop. The descriptions could be done that way probably as well to some extent. If you don't want to do it, I can accept this and do it later.

white-haired-uncle:
I prefer the table for a small solution set (and I expect it to stay small, I've found by redeem level 10 or so I no longer use redeem to try to unlock advancements I'll never earn the XP to use anyway) vs a formula that is a bit interesting.  It's also easier to to make arbitrary changes (why?), and easier to see anomolies (like L11 vs L12).  I thought about putting the formula in comments, but then decided it was complex enough just describing it would be over half the work of implementing it.

I also considered using the function, however the data set was generated,  to auto-populate the descriptions.  That's the main reason I have a separate lua function for the hit chance.  But it's a small set, unlikely to change, and I kind of hate doing that kind of thing (I like the end result, but the coding is not fun).

Toranks:
Don't forget to include this, or redeem will be difficult for the player to see when is recharged:

https://github.com/Dugy/Legend_of_the_Invincibles/pull/557/files

white-haired-uncle:
Thanks.  That's a nice addition.  I play on ludicrous speed, so I usually don't see messages like that.

I wonder if there's a way I can get git to just send diffs in case the file I've been working on has changed.  It seems to be sending the whole file.  I only want to send the diffs against the copy I was working on, not replace everything.

white-haired-uncle:
Added a reference in lua/redeem.lua comments to the WML redeem formulae

PR #612, opened 2023-07-20
replace FOREACH

white-haired-uncle:
Tested the 5 scenarios that I could.

Akula's comment in Diplomacy about not being able to swim is a little odd.  There's only a bit of deep water in an unused corner (it would make more sense if the river was deep water).  And of course, I immediately gave her Static Float, and got not so much as a thank you.  I don't think she's a very nice person.

In Lilith, Phobos speaks (giving away his location), though he's invisible

In Lilith, this does not work: (changed to id=)
speaker=Verderber,Nemesis,Phobos,Deimos

In Lilith, comment says when 3 of the councilors die.  It only takes 2 if they die before Corruptia.

### DIFF
--- a/scenarios2/10_Lilith.cfg
+++ b/scenarios2/10_Lilith.cfg
@@ -1026,9 +1026,9 @@
                         [set_specials]
                             mode=append
                             {WEAPON_SPECIAL_MAGICAL}
-                            #textdomain wesnoth-units
+#textdomain wesnoth-units
                             {WEAPON_SPECIAL_PLAGUE_TYPE_LOTI Wraith _"Wraith"}
-                            #textdomain wesnoth-loti
+#textdomain wesnoth-loti
                         [/set_specials]
                     [/effect]
                     [effect]
@@ -1086,9 +1086,9 @@
                         [set_specials]
                             mode=append
                             {WEAPON_SPECIAL_MAGICAL}
-                            #textdomain wesnoth-units
+#textdomain wesnoth-units
                             {WEAPON_SPECIAL_PLAGUE_TYPE_LOTI Shadow _"Shadow"}
-                            #textdomain wesnoth-loti
+#textdomain wesnoth-loti
                         [/set_specials]
                     [/effect]
                     [effect]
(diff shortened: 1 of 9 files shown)


scenarios2/10_Lilith.cfg
```diff
@@ -1083,9 +1086,9 @@
                         [set_specials]
                             mode=append
                             {WEAPON_SPECIAL_MAGICAL}
-#textdomain wesnoth-units
+                            #textdomain wesnoth-units
                             {WEAPON_SPECIAL_PLAGUE_TYPE_LOTI Shadow _"Shadow"}
-#textdomain wesnoth-loti
+                            #textdomain wesnoth-loti
```
Dugy, line 1091: Unintended indentation probably.

Dugy (changes requested):

Dugy:
> Akula's comment in Diplomacy about not being able to swim is a little odd.

It's just foreshadowing about the fact that she a lot of steel machinery in her body. It's not consequential for anything because almost no units can pass through deep water anyway.

> In Lilith, Phobos speaks (giving away his location), though he's invisible

He can be heard talking without being seen, no?

> In Lilith, comment says when 3 of the councilors die. It only takes 2 if they die before Corruptia.

Figuring out a wording that would be accurate without being overcomplicated is no easy task.

Otherwise we might end up with wordings like this one in Path of Exile: _50% chance to Trigger Socketed Spells when you Spend at least 100 Mana on an Upfront Cost to Use or Trigger a Skill, with a 0.1 second Cooldown._

white-haired-uncle:
> > In Lilith, Phobos speaks (giving away his location), though he's invisible
> 
> He can be heard talking without being seen, no?

When he speaks, there's a white hex shaped border around the tile where he's "hiding".  So even though you can't see him, you've been handed his location.

I doubt it matters as he usually attacks as soon as you come in range anyway, but it looked odd enough to warrant a mention.

Dugy:
> When he speaks, there's a white hex shaped border around the tile where he's "hiding".

It can be fixed by checking for the circumstances where this happens and have the narrator speak with his name as caption and his portrait as picture, but personally I would not bother.

white-haired-uncle:
That sounds like a great exercise for somebody else.

Dugy:
Okay.

PR #613, opened 2023-07-21
Trader

white-haired-uncle:
Added cost in item_list.cfg for each item that was defined at trader
Added wml_var in item_list.cfg for gems (e.g. for item.name=Ruby you can now use wml_var to access global "rubies")

Fix for items.lua ("fix" depending on how you define "between", I assume inclusive, one could make a case for exclusive, but mixed is right out IMO)

New trader.lua (probably temporary) contains trader code

For testing, updated scenarios4/02
     - enter village at 21,29 to visit trader
     - enter village at 16,26 to reload trader.lua w/o restarting scenario
     - included tests/saves/LotI1-Into\ Our\ Frozen\ Land.gz for testing

TODO:
     - Add confirmation of item purchase
     - MP stuff
     - Figure out why buying items works differently (wrong) the first time only


lua/trader.lua
```diff
@@ -0,0 +1,312 @@
...
+				wml.tag.column {
+					horizontal_grow = true,
+					wml.tag.toggle_panel { listbox_template }
+				}
+			}
+		}
+	}
+
+	-- 3) Top-level dialog, includes help message and listbox.
+	local query={_"What would you like to buy?", _"Buy something or get out", _"See anything to your liking?", _"I have a few items here that fell off a wagon"}
```
Dugy, line 131: I love the last option.

white-haired-uncle:
I'm not having much luck with wml.fire("item_pick",{???}) but I think once I figure it out it might work better than wml.tag.fire_event when I want to actually pick something up.

white-haired-uncle:
I still need to iron out how the gems will work.  Might as well just treat them like other items, but keep a local copy of the count, so the "currently have X of them" does get updated immediately.

And I should add a message showing current gold that accounts for purchases now, even though global gold won't update right away.

Added offer field to item_list.cfg for gems (e.g. "Buy an emerald for").  Initially I was determining whether to use a/an by the first letter of the gem name, then realized not everyone speaks en_US.

I think the MP safe issue is pretty much the last straw.

Stuck with the ok button vs buy.  There's something automagical about the ok button that makes things just work.  I buy button could probably be made to work, but it's too much work for now (if I ever do the autorecall work I want to I'll have to figure it out then and I came come back to this).

Didn't do anything with random items.  [trader_menu] takes a list of items as input, one could always select random items to include there, for now at least.

[trader_menu] takes a unique id, which is used to record purchased items, so you can only purchase a non-gem item (e.g. Holy Sword) once from each trader, even if you leave the scenario and come back or something.

Changed the confirmation view around a bit.  I like this better.
Screenshot from 2023-07-22 12-30-54
Screenshot from 2023-07-22 12-26-44

Dugy:
I wonder that maybe the gems could be in a secondary window opened by a button (something like _Gem trade_), or a separate column. Items might have their descriptions right below the names, it would be helpful for players who don't know them from memory (which is probably most players).

That vendor's name, _Owreorer,_ sounds so lame. Right now I am a little ashamed I had written the code that generates names in Wesnoth.

white-haired-uncle:
> I wonder that maybe the gems could be in a secondary window opened by a button (something like _Gem trade_), or a separate column. 

That might look a bit odd, particularly if you buy all the other items.

> Items might have their descriptions right below the names, it would be helpful for players who don't know them from memory (which is probably most players).

How about a popup when you mouse over the item?  (You can't see my mouse, but it's pointing at "Holy Sword".  If I point at the HS's image, it pops up the flavour).

Screenshot from 2023-07-23 02-18-16

Dugy:
> That might look a bit odd, particularly if you buy all the other items.

And if these two were separate columns?

> How about a popup when you mouse over the item?

Yeah, that sounds better.

white-haired-uncle:
> > That might look a bit odd, particularly if you buy all the other items.
> 
> And if these two were separate columns?
> 

Assuming it wouldn't show an empty column, I'm still not a fan, but I can see where it could be nice if playing on a tablet where horizontal space is limited?

But, I think you'd then be dealing with 2 listboxes.  Perhaps not, but I've had just about enough of this simple exercise.  I am considering something like that if I ever do a new autorecall menu (one column is autorecall list, one column is free list), so if I have to figure it out then I could always come back to this if it's worth doing then.  I think it's worked well enough so far (in WML) the way it is. And if space is an issue, there's probably gains to be had by messing around with borders, further cropping the gem images, or just getting rid of the gem images (I like them, but they're too small for my taste anyway).

white-haired-uncle:
A little cleanup necessary.  I'd like to get rid of the recursion and just use a while loop.  Not really worth doing as long as the MP issues remain.

white-haired-uncle:
Update to description of Tome of Liches

Replace Nothing option in item selection with Ok/Close

Replace Yes/No/Ok in confirmation with Yes/No buttons

New attempt at the MP stuff

Dugy:
I tried to run it to help you with the synchronisation problem, but I got a big error that _Grid row 2 must have at least one column.__

white-haired-uncle:
I don't get it.  I tried to upload another copy of the lastest version, which I've tested way too much, but it says there's no changes.

What do you get for cksum lua/trader.lua ?
1132261609 11812 lua/trader.lua

Dugy:
I get `3264902517 12997 lua/trader.lua`

white-haired-uncle:
See previous comment I was just editing.

Dugy:
Yeah, it works at f51b42a.

lua/items.lua
```diff
@@ -196,7 +196,7 @@ loti.item.type.numbers_between = function(start_number, end_number)
 	local numbers = {}
 	for item_number in pairs(loti.item.type._load()) do
 		if not start_number or item_number >= start_number then
-			if not end_number or item_number < end_number then
+			if not end_number or item_number <= end_number then
```
edwardspec, line 199: If you are changing this, you should also update the invocation of `numbers_between()` in `lua/crafting.lua`, which currently assumes that `end_number` is excluded.

lua/trader.lua
```diff
@@ -0,0 +1,406 @@
+--! #textdomain "wesnoth-loti"
+local _ = wesnoth.textdomain "wesnoth-loti"
+
+local function confirm_purchase(item,trader)
+
+	-- Top-level dialog, includes help message and listbox.
+	local confirm_queries = { _"A bargain at twice the price.  We good?", _"Makes a great gift.  Wrap it up to go?", _"You gonna buy that or what?",
+		_"I close in 5 minutes.  We got a deal or what?", _"Look at it.  Do you really want it?" }
+	local confirm_query=confirm_queries[math.random(1,#confirm_queries)]
```
edwardspec, line 9: code style: need spaces around `=` and after `,`
(here and in other places)

lua/trader.lua
```diff
@@ -0,0 +1,406 @@
...
+							if (bought_it) then return end
+						end
+					end
+					if (debug) then print(string.format("id = %d, name = %s, cost = %d, image = %s, desc = %s, flav = %s",
+						id,item.name,item.cost,item.image,item.description,item.flavour)) end
+
+					local label = item.name .. " (" .. item.cost .. _ " gold" .. ")"
+					if (item.sort == "gem") then
+						local have = gem_count[item.number]
+						if (have == nil) then have=0 end
```
edwardspec, line 182:
```lua
local have = gem_count[item.number] or 0
```

lua/trader.lua
```diff
@@ -0,0 +1,406 @@
...
+						label=item.offer .. " " .. tostring(item.cost) .. _ " (currently have " .. have .. _ " of them)"
+					end
+
+					local allowed = true
+					if (item.cost > gold) then allowed = false end
```
edwardspec, line 187:
```lua
local allowed = item.cost <= gold
```

lua/trader.lua
```diff
@@ -0,0 +1,406 @@
...
+			end
+
+			local selected_menu_item = nil
+			local function postshow(dialog)
+				selected_menu_item = get_selected_menu_item(dialog)
+			end
+
+			gui.show_dialog(dialogDefinition, preshow, postshow)
+
+			if (done) then return end  -- pressed the close button
```
edwardspec, line 346: don't need `(` and `)`, here and in other places

lua/trader.lua
```diff
@@ -0,0 +1,406 @@
...
+		return { value=stuff }
+		end
+	)
+	local items={}
+	for _,list in pairs(bought_items) do
+		for w in string.gmatch(tostring(list), "[%d]+") do table.insert(items,tonumber(w)) end
+	end
+	for _,item_num in pairs(items) do
+		local item=loti.item.type[item_num]
+		wml.fire("gold",{side=side_number,amount=-item.cost})
```
edwardspec, line 402: Should directly modify `side.gold` (integer).

lua/trader.lua
```diff
@@ -0,0 +1,406 @@
...
+	local ret_val=false
+
+	local function preshow(dialog)
+		dialog.ok.on_button_click = function()
+			ret_val = true
+		end
+	end
+
+	local function postshow()
+	end
```
edwardspec, line 124: This empty function isn't needed: `postshow` argument of `show_dialog()` is optional.

edwardspec (changes requested):

PR #615, opened 2023-07-24
add EVASION macro

white-haired-uncle:
Move some redundant code into a single macro.

Originally, there was one taunt for the first time (unless you got extremely lucky), and then one for subsequent misses.  I broke that, could fix it, but I don't think it's worth it (I was thinking you pass a message when you want a specific message, or "" to let the macro pick one for you).

I put a variable in a message.  I really like it.  I can take it out.

It looks like only one attacker_misses is getting used.  Reversing the order (so the one time taunt comes first) was the only way to get the taunt event to run (other than commenting out the other one).  This is consistent with the previous version, but probably not what was originally intended.

### DIFF
--- a/scenarios7/14_Arctic_Wastelands.cfg
+++ b/scenarios7/14_Arctic_Wastelands.cfg
@@ -962,7 +962,7 @@ You can build a keep anywhere for 50 gold, or a village for 10 gold. (see option
             experience=no
         [/kill]
     [/event]
-    {EVASION lethalia_evasion "Xuc,Xu,Xol,Xos,Zic,Ms^Xm"}
+    {EVASION lethalia_evasion}
     [event]
         name=attacker_hits
         [filter_second]
--- a/scenarios7/14_Arctic_Wastelands.cfg
+++ b/scenarios7/14_Arctic_Wastelands.cfg
@@ -962,47 +962,7 @@ You can build a keep anywhere for 50 gold, or a village for 10 gold. (see option
             experience=no
         [/kill]
     [/event]
-    [event]
-        name=attacker_misses
-        first_time_only=no
-        [filter_second]
-            ability=lethalia_evasion
-        [/filter_second]
-        [store_locations]
-            [filter_adjacent_location]
-                x,y=$x2,$y2
-            [/filter_adjacent_location]
-            [not]
-                terrain=Xuc,Xu,Xol,Xos,Zic,Ms^Xm
-            [/not]
-            variable=dodge_locs
-        [/store_locations]
-        [foreach]
-            array=dodge_locs
-            [do]
-                [if]
-                    [have_unit]
-                        x,y=$this_item.x,$this_item.y
-                    [/have_unit]
-                    [else]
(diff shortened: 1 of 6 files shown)


scenarios7/14_Arctic_Wastelands.cfg
```diff
@@ -962,47 +962,7 @@ You can build a keep anywhere for 50 gold, or a village for 10 gold. (see option
...
-        name=attacker_misses
-        [filter_second]
-            ability=lethalia_evasion
-        [/filter_second]
-        [message]
-            speaker=second_unit
-            message= _ "Hahaha, you still think you can match my dexterity?"
-        [/message]
-    [/event]
+    {EVASION lethalia_evasion "Xuc,Xu,Xol,Xos,Zic,Ms^Xm"}
```
Dugy, line 965: Can you just put all these hex types into the event and don't pass it as an argument? I think the extra terrains that would match the filter would not change anything.

utils/utils.cfg
```diff
@@ -1837,3 +1837,89 @@ $item_info.description"
...
+        [set_variables]
+            name=msgs
+            [value]
+                message=_"Hahaha, you think you can match my dexterity, $unit.name|?"
+            [/value]
+            [value]
+                message=_"Hahaha, nobody can match my dexterity!"
+            [/value]
+            [value]
+                message=_"Missed me by that <span font='oblique'>that</span> much"   # TV show Get Smart
```
Dugy, line 1857: I would not bother with the comments where do these quotes come from, but if you want...

scenarios7/14_Arctic_Wastelands.cfg
white-haired-uncle: Yes, but Uria's look pretty different from the rest.  Would "Ms^Xm,Q*,X*,Zic" work for everything?

utils/utils.cfg
white-haired-uncle: I put those in there for translators, in case a phrase could be translated more than one way and they didn't catch the reference.

scenarios7/14_Arctic_Wastelands.cfg
Dugy: Yes, I am quite sure these terrains don't appear around the evil Lethalia.

utils/utils.cfg
Dugy: Except for film references, they would not be able to use the references in the target languages. They'd have to find some other references instead.

white-haired-uncle: True, but at least I've attempted to convey what the quote refers to in en_US.  For example, there's a special taunt when Efraim misses L_evil, which I picked due to their prior dialog.  It's not important to translate verbatim, but to understand why she's saying what she's saying.

If it doesn't translate and/or they come up with a replacement that has no relation to the original reference, that's fine IMO.  And even if I didn't like it, I won't be able to read it anyway.

Dugy: All right.

PR #616, opened 2023-07-25
gold carryover scenarios1

white-haired-uncle:
20230725 10:17:23 warning deprecation: NEW_GOLD_CARRYOVER_NOTE_40 has been deprecated and may be removed in version 1.17.0.; This is now automatically generated and not required. Use [objectives][gold_carryover]carryover_percentage=40 instead.
20230725 10:17:23 warning deprecation: EARLY_FINISH_BONUS_NOTE has been deprecated and may be removed in version 1.17.0.; This is now automatically generated and not required. Use [objectives][gold_carryover]bonus=yes instead.

Breaking up by chapter for easy review.

I'd like to see these in every scenario, instead of assuming player knows the defaults (which could change).

### DIFF
--- a/scenarios2/05_Umbra.cfg
+++ b/scenarios2/05_Umbra.cfg
@@ -217,9 +217,9 @@
                         [set_specials]
                             mode=append
                             {WEAPON_SPECIAL_POISON}
-#textdomain wesnoth-units
+                            #textdomain wesnoth-units
                             {WEAPON_SPECIAL_PLAGUE_TYPE_LOTI Shadow _"Shadow"}
-#textdomain wesnoth-loti
+                            #textdomain wesnoth-loti
                         [/set_specials]
                     [/effect]
                 [/object]
@@ -319,7 +319,10 @@
                 description= _ "Turns run out"
                 condition=lose
             [/objective]
-            note={EARLY_FINISH_BONUS_NOTE}+{NEW_GOLD_CARRYOVER_NOTE_40}
+            [gold_carryover]
+                carryover_percentage=40
+                bonus=yes
+            [/gold_carryover]
         [/objectives]
     [/event]
 
--- a/scenarios2/05_Umbra.cfg
+++ b/scenarios2/05_Umbra.cfg
@@ -217,9 +217,9 @@
                         [set_specials]
                             mode=append
                             {WEAPON_SPECIAL_POISON}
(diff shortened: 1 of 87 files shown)


white-haired-uncle:
Two hundred scenarios, and one comment...

```
#No idea what the hell is this - but it may have some purpose. Was I drunk when
I was writing this?
```

white-haired-uncle:
Okay, that was a little messy, but part1 is now free of gold carryover messages.

Dugy:
Can I merge it now?

white-haired-uncle:
sure

scenarios2/05_Umbra.cfg
```diff
@@ -217,9 +217,9 @@
                         [set_specials]
                             mode=append
                             {WEAPON_SPECIAL_POISON}
-#textdomain wesnoth-units
+                            #textdomain wesnoth-units
```
Dugy, line 220: Unwanted change.

scenarios2/05_Umbra.cfg
```diff
@@ -217,9 +217,9 @@
...
                             {WEAPON_SPECIAL_PLAGUE_TYPE_LOTI Shadow _"Shadow"}
-#textdomain wesnoth-loti
+                            #textdomain wesnoth-loti
```
Dugy, line 222: Unwanted change.

Dugy (changes requested):

white-haired-uncle:
Just confirmed it's wmlindent doing that.

scenarios2/05_Umbra.cfg
white-haired-uncle: done

scenarios2/05_Umbra.cfg
white-haired-uncle: done

white-haired-uncle:
And that's the last of them (not incl spinoffs).

I did notice there are some scenarios in ch10 (at least?) where the gold carry over is only set (to 100) at victory.  I believe this would be misleading if the player assumes that no info in objectives means the wesnoth default.

Dugy:
> I believe this would be misleading if the player assumes that no info in objectives means the wesnoth default.

Yes, it could be misleading, it does make sense to add carryover notes about that.

white-haired-uncle:
yes

PR #617, opened 2023-07-26
a little easier

white-haired-uncle:
While testing, I played 10/13 Black Souls.  It was as bad as I remember.  Choosing/crafting just the right gear, moving/attacking very very carefully, save loading a couple times, taking about 4 greater healing potions, and using skirmisher to race through the last 20% or so of the map, I barely made it in time.  Having a anti-demon potion was probably the difference between success and failure.

Propose a few tweaks.  It's still going to be quite hard, but it is the last real scenario.  I tried these on normal, and I found the changes to be just enough to give you a fair chance of surviving the first few turns without cheating (granted, I only played it once, and only for the first several turns).

- Added two turns
- Decreased number of black souls at beginning on easy
- For easy/normal, move black souls back slightly, giving player a chance to live through turn 2

### DIFF
--- a/scenarios10/13_Black_Souls.cfg
+++ b/scenarios10/13_Black_Souls.cfg
@@ -5,7 +5,7 @@
     map_data="{~add-ons/Legend_of_the_Invincibles/maps/13_Laboratory_Centre.map}"
     next_scenario=14_End_of_the_World
     {GLOBAL_EVENTS}
-    {TURNS 25 23 21}
+    {TURNS 27 25 23}
     experience_modifier=80
     victory_when_enemies_defeated=no
     {INDOORS}
@@ -150,6 +150,9 @@
         [/recall]
         {PLACE_IMAGE scenery/portal-red.png 1 30}
         {VARIABLE spawn_count 30}
+#ifdef EASY
+        {VARIABLE spawn_count 25}
+#endif
         [while]
             [variable]
                 name=spawn_count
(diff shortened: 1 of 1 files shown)


Dugy (approved):

Dugy:
I am fine with the proposed changes, I will accept it as soon as you mark it as ready.

Dugy:
Are you sure you wanted to close this?

white-haired-uncle:
No.  I had some commits that wouldn't go away so I ended up nuking the repo.

Dugy:
I was usually fixing this with `git reset --hard some_ref` and cherrypicking the commits I wanted, but this also works.

white-haired-uncle:
Not sure how to do that 
- identify some_ref
- remotely or from github

I'm sure git is a heck of a tool, but it's got a bit of a learning curve (more than I want to take on at one time)

Dugy:
Yeah, it takes a while to learn git to the point of being able to fix things that are broken.

In this case, you probably want to start off some commit or the top of some branch, most likely master, and add some commits on top of that. If you want to start from master, you write this into your local clone, using CLI:

`git reset --hard master`

This makes your branch's content identical to master. Then you look at the hashes of commits you want to add, like 44d505a, and cherrypick them:

`git cherry-pick 44d505a`

You do this for every commit.

Then you can make these changes to the remote of your branch as well:

`git push --force`

PR #618, opened 2023-07-26
Abilities

white-haired-uncle:
I have an enhancement request to improve how wesnoth handles tooltips for units with lots of items. 

https://forums.wesnoth.org/viewtopic.php?p=682946#p682946

Of course, we can always go to Unit Information, but I'd prefer to do that only when necessary.  To make it more likely that the abilities will fit in the main wesnoth gui tooltip, I propose shortening some of the descriptions.

I'd also like to figure out where wesnoth adds blankline, ----------, blankline to the list and get rid of those.


Dugy:
That makes sense.

You have accidentally put an unrelated commit into the branch, so I manually added 07781c1 (the commit you wanted to put here) into master.

PR #619, opened 2023-07-26
fix crafting side-effect

white-haired-uncle:
When I fixed loti.item.type.numbers_between, I introduced a side-effect in crafting.

This time I did a search, and nothing else should have been affected.

I still don't like the whole idea of that function.  If you look at item_list.cfg there's a comment that Dugi's Wrath will  have the highest number.  Next you see several items with higher numbers.  Because things change.  Oh well, I used it myself, so who am I to complain?

### DIFF
--- a/lua/crafting.lua
+++ b/lua/crafting.lua
@@ -599,7 +599,7 @@ loti.gem.show_crafting_window = function(x, y)
 					dialog.gui_recipe_chosen:remove_items_at(1, 0)
 
 					order = 1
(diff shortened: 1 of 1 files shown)


Dugy:
Off by one error, right?

Dugy:
Well, I am used to seeing a range defined by the lowest number included and the number just above the maximum, and all the lua indexing is quite counter-intuitive to me. I assume you're having a similar problem.

Issue #621, opened 2023-07-26
TBC textdomains


white-haired-uncle:
```
spartan$ grep -r wesnth-tbc The_Beautiful_Child/
The_Beautiful_Child/utils/rpg_events.cfg:#textdomain wesnth-tbc
The_Beautiful_Child/utils/utils.cfg:#textdomain wesnth-tbc
The_Beautiful_Child/utils/global_events.cfg:#textdomain wesnth-tbc
The_Beautiful_Child/utils/strategy_events.cfg:#textdomain wesnth-tbc

spartan$ grep -r wesnoth-tbc The_Beautiful_Child/
The_Beautiful_Child/scenarios/07_Siege_of_Ziggurat.cfg:#textdomain wesnoth-tbc
The_Beautiful_Child/scenarios/03_To_Tame_a_Land.cfg:#textdomain wesnoth-tbc
The_Beautiful_Child/scenarios/08_Bringing_the_Light.cfg:#textdomain wesnoth-tbc
...
The_Beautiful_Child/units/Exalted_Blackguard.cfg:#textdomain wesnoth-tbc
The_Beautiful_Child/_main.cfg:#textdomain wesnoth-tbc
The_Beautiful_Child/_main.cfg:    name="wesnoth-tbc"
```

I don't know anything about textdomains, but those first four are typos, correct?  I assume if they were correct there would be something to "define" them like wesnoth-tbc has?

EDIT:  Actually, now that I look at the diffs, it looks like wmlindent did that.  I suppose I should set them back?  And is that a bug in wmlindent that needs to be reported?  thx

Dugy:
I think it's a typo, there is no reason for anything to be named `wesnth`.

Toranks:
I already fixed that on another PR, pls update your code regulary

white-haired-uncle:
Hmm, I updated yesterday or maybe the day before.

Maybe I had a shell in the old directory.

PR #622, opened 2023-07-26
replace FOREACH in spinoffs

white-haired-uncle:
Looks like wmlindent had a lot to do here.  The only changes I made were to replace the FOREACH loops with for/foreach, and I changed the loop variables in a couple cases that used i for both inner and outer loops (not wrong as I understand it, but IMO it's a lot more readable to just use unique and potentially less error-prone).

I haven't tested any of these.  I'm going to do some other cleanup stuff (have these been brought up to 1.16?  a recent forum comment said no) and then play test.

Ref #621

### DIFF
--- a/spinoffs/The_Beautiful_Child/units/Bringer_of_Light2.cfg
+++ b/spinoffs/The_Beautiful_Child/units/Bringer_of_Light2.cfg
@@ -1,5 +1,4 @@
-#textd
-main wesnoth-tbc
+#textdomain wesnoth-tbc
 #define DAZZLE_COUNTER
     [chance_to_hit]
         id=dazzle_counter
--- a/spinoffs/The_Beautiful_Child/units/Bringer_of_Light2.cfg
+++ b/spinoffs/The_Beautiful_Child/units/Bringer_of_Light2.cfg
@@ -1,20 +1,21 @@
-#textdomain wesnoth-tbc
+#textd
+main wesnoth-tbc
 #define DAZZLE_COUNTER
-[chance_to_hit]
-    id=dazzle_counter
-    value=30
-    apply_to=opponent
-    [filter_base_value]
-        greater_than=30
-    [/filter_base_value]
-    [filter_opponent]
-        [filter_wml]
-            [variables]
-                dazzled=yes
-            [/variables]
-        [/filter_wml]
-    [/filter_opponent]
-[/chance_to_hit]
+    [chance_to_hit]
+        id=dazzle_counter
+        value=30
+        apply_to=opponent
+        [filter_base_value]
+            greater_than=30
+        [/filter_base_value]
+        [filter_opponent]
+            [filter_wml]
+                [variables]
+                    dazzled=yes
+                [/variables]
+            [/filter_wml]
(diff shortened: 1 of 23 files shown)


Dugy:
Can you please test at least a little of some parts? It's not possible to review it because wmllindent has changed everything.

spinoffs/The_Beautiful_Child/utils/utils.cfg
```diff
@@ -1,4 +1,4 @@
-#textdomain wesnoth-tbc
+#textdomain wesnth-tbc
```
edwardspec, line 1: typo

spinoffs/The_Beautiful_Child/utils/global_events.cfg
```diff
@@ -1,51 +1,53 @@
-#textdomain wesnoth-tbc
+#textdomain wesnth-tbc
```
edwardspec, line 1: typo

spinoffs/The_Beautiful_Child/units/Bringer_of_Light2.cfg
```diff
@@ -1,20 +1,21 @@
-#textdomain wesnoth-tbc
+#textd
+main wesnoth-tbc
```
edwardspec, line 2: typo

edwardspec (changes requested):

edwardspec:
> It's not possible to review it because wmllindent has changed everything.

Can add `?w=1` to the URL of GitHub diff to ignore whitespace changes: https://github.com/Dugy/Legend_of_the_Invincibles/pull/622/files?w=1

Dugy:
Oh thanks. It looks okay to me. Except for the textdomain issues.

spinoffs/The_Beautiful_Child/utils/utils.cfg
white-haired-uncle: done

spinoffs/The_Beautiful_Child/utils/global_events.cfg
white-haired-uncle: done

spinoffs/The_Beautiful_Child/units/Bringer_of_Light2.cfg
white-haired-uncle: done

edwardspec (approved):

spinoffs/The_Beautiful_Child/utils/utils.cfg
edwardspec: Still exists.

white-haired-uncle:
In Kill the King, it always says there's an early finish bonus, but it's 0 per turn.  I think this  is because gold/village and base income are 0.

white-haired-uncle:
> Can you please test at least a little of some parts? It's not possible to review it because wmllindent has changed everything.

I played all of KtK, not realizing the only foreach was at the very end of the last scenario.  And it was broken ([do] before array=).

PR #623, opened 2023-07-28
update ktk

white-haired-uncle:
Update difficulty selection.  Needs better names (groupie, musician, rock star was the best I could come up with).

Update gold.  I don't like the bonus=yes, but no bonus (tell player there's early finish bonus, but it's 0).

Add [/unit_type] to end of most custom units.  I don't know that that comment at the end means, but it wouldn't load w/o this.

[advance] => [advancement] on one of the units.

Played on 1.16 okay.

### DIFF
--- a/spinoffs/Kill_the_King/_main.cfg
+++ b/spinoffs/Kill_the_King/_main.cfg
@@ -14,8 +14,28 @@
 
     abbrev= _ "KtK"
     rank=9000
-    difficulties=EASY,NORMAL,HARD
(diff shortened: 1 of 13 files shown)


Dugy (approved):

Dugy:
>  Needs better names (groupie, musician, rock star was the best I could come up with).

Maybe __roadie__ for medium difficulty would be better, otherwise I like the suggested names.

> Update gold. I don't like the bonus=yes, but no bonus (tell player there's early finish bonus, but it's 0).

Yeah, that makes sense.

white-haired-uncle:
yeah, roadie is what i meant for easy, not groupie.

white-haired-uncle:
BTW, the drummer -- reference to Spoonman?

Dugy:
> BTW, the drummer -- reference to Spoonman?

No, I don't know who is Spoonman.

white-haired-uncle:
It's an old Soundgarden tune

"And all my friends are skeletons,
they beat the rhythm with their bones"

Dugy:
Definitely a coincidence.

PR #624, opened 2023-07-28
tbc updates for 1.16

white-haired-uncle:
Change advance -> advancement

In Crooked Ravine:
    - Remove she_brother from brothers_store when killed/defected
    - Make most of brothers appearance dialog one time only
    - rename quest variable for consistency
    
Limit main character to one legacy (was two)

Add Konrad II to your party when found

One of the mercs was missing payment when recruited

In Fertile Lands, units killed by secondary effect like that from shockwave interrogate themself, skip that

In Valley Passage, after killing Luke, units that couldn't traverse mountains couldn't traverse mountains.

In final scenario, enemies may block stairway (specifically, if you enter the final room and then leave, they will just keep spawning new units, which does not honor the [ai][avoid]), blocking you from re-entering room.
    - Added [avoid], may not be needed
    - Added turn end event so that units spawned on top of stairs (and leader may move there ignoring avoid, probably trying to get to a keep -- should that tile be a keep?) will move.
    - This works until enemy has 50 or so units in room7.  Changed BoL2 to only cast spells when side 1 is visible.  They still recruit, which will probably block the stairs at some point, but it's only 1 unit per turn.  Could/should move the keeps.

"error wml: [unstore_unit]: variable 'undazzled' doesn't exist" - might be debug-only
    - changed dazzle to blind.  untested other than it loads

TODO:
warning wml: Couldn't find right animation for unit player  (he's a Duke)

info deprecation: [set_specials]mode=<unset> has been deprecated indefinitely.; The mode defaults to 'replace', but should often be 'append' instead. The default may change in a future version, or the attribute may become mandatory.

Balancing:  Playing on normal, I find this super hard at first (I have to constantly run away using obelisk until I get enough advancement to fight).  Then by about Fertile Lands I could probably just about walk through with 3 units (I suppose it's fun for people who just want to collect tons of items and explore AMLAs), and the merchants before the final scenario are completely unnecessary.  Also, I'm not sure there's much difference in levels.  Thoughts:
  - Change the  drop frequency by difficulty (and turn it down by at least 25% on normal, likely a lot more)
  - When you fight Luke, he gets crippled if your party is "small".  Either don't do that on hard, or change the limits.
  - Gold starts off pretty hard to get (which is okay, not much to spend it on other than a couple scenarios where you need fodder), but by the middle I've got >1500 more than I'll ever need.

### DIFF
--- a/spinoffs/The_Beautiful_Child/scenarios/Valley_Passage.cfg
+++ b/spinoffs/The_Beautiful_Child/scenarios/Valley_Passage.cfg
@@ -73,13 +73,13 @@
                     name= _"Konrad II"
                     canrecruit=yes
                     [modifications]
-                        [advance]
+                        [advancement]
                             id=portrait
                             [effect]
                                 apply_to=profile
                                 portrait="portraits/humans/transparent/fencer.png"
                             [/effect]
-                        [/advance]
+                        [/advancement]
                     [/modifications]
                 [/unit]
             [/else]
--- a/spinoffs/The_Beautiful_Child/scenarios/Valley_Passage.cfg
+++ b/spinoffs/The_Beautiful_Child/scenarios/Valley_Passage.cfg
@@ -283,12 +283,7 @@
                         radius=7
                         side=1
                     [/remove_shroud]
-                    [move_unit]
-                        side=1
-                        to_x=17
-                        to_y=12
-                        fire_event=no
-                    [/move_unit]
+                    {MOVE_SIDE_TO 1 17 12}
                     [message]
                         speaker=Luke
                         message=_"Work, dig in the ground, the ancient vault must be somewhere around here. If I was Niflheim the Horror from the Northern Wastelands, I would hide it somewhere around here."
@@ -890,42 +885,55 @@
                         [store_unit]
                             [filter]
                                 side=1
-                                [filter_wml]
-                                    [movement_costs]
-                                        mountains={UNREACHABLE}
-                                    [/movement_costs]
-                                [/filter_wml]
                             [/filter]
-                            variable=cant_pass
+                            variable=side1
                         [/store_unit]
+                        {VARIABLE changed_mnt_cost no}
                         [for]
-                            array=cant_pass
+                            array=side1
                             [do]
-                                [set_variables]
-                                    name=cant_pass[$i].modifications.advancement[$cant_pass[$i].modifications.advancement.length]
-                                    mode=replace
-                                    [value]
-                                        id=mountain_passing
-                                        [effect]
-                                            apply_to=movement_costs
-                                            replace=true
-                                            [movement_costs]
-                                                mountains=3
-                                            [/movement_costs]
-                                        [/effect]
-                                    [/value]
-                                [/set_variables]
-                                {VARIABLE cant_pass[$i].movement_costs.mountains 3}
-                                [unstore_unit]
-                                    variable=cant_pass[$i]
-                                    find_vacant=no
-                                [/unstore_unit]
+                                [if]
+                                    [variable]
+                                        name=side1[$i].movement_costs.mountains
+                                        greater_than=0    # not null
+                                    [/variable]
+                                    [and]
+                                        [variable]
+                                            name=side1[$i].movement_costs.mountains
+                                            less_than=4
+                                        [/variable]
+                                    [/and]
+                                    [else]
+                                        {VARIABLE changed_mnt_cost yes}
+                                        {VARIABLE helped $side1[$i].id}
+                                        [set_variables]
+                                            name=side1[$i].modifications.advancement[$side1[$i].modifications.advancement.length]
+                                            mode=replace
+                                            [value]
+                                                id=mountain_passing
+                                                [effect]
+                                                    apply_to=movement_costs
+                                                    replace=true
+                                                    [movement_costs]
+                                                        mountains=3
+                                                    [/movement_costs]
+                                                [/effect]
+                                            [/value]
+                                        [/set_variables]
+                                        {VARIABLE side1[$i].movement_costs.mountains 3}
+                                        [unstore_unit]
+                                            variable=side1[$i]
+                                            find_vacant=no
+                                        [/unstore_unit]
+                                    [/else]
+                                [/if]
                             [/do]
                         [/for]
+                        {CLEAR_VARIABLE side1}
                         [if]
                             [variable]
-                                name=cant_pass.length
-                                greater_than=0
+                                name=changed_mnt_cost
+                                equals=yes
                             [/variable]
                             [then]
                                 [message]
@@ -933,7 +941,7 @@
                                     message= _ "I think that we'll need a heavy hike through the mountains, can all of us pass through mountains?"
                                 [/message]
                                 [message]
-                                    id=$cant_pass.id
+                                    id=$helped
                                     message= _ "I hope I'll manage."
                                 [/message]
                                 [message]
@@ -948,7 +956,7 @@
                                 [/message]
                             [/else]
                         [/if]
-                        {CLEAR_VARIABLE cant_pass}
+                        {CLEAR_VARIABLE changed_mnt_cost,helped}
                         [message]
                             speaker=Sicaria
                             message= _ "And we'll see Konrad then, right?"
--- a/spinoffs/The_Beautiful_Child/scenarios/Valley_Passage.cfg
+++ b/spinoffs/The_Beautiful_Child/scenarios/Valley_Passage.cfg
@@ -902,7 +902,7 @@
                             array=cant_pass
                             [do]
                                 [set_variables]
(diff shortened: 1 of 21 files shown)


white-haired-uncle:
Does this: sicaria_store.modifications.advance[$sicaria_store.modifications.advance] need to use advancement as well?

Dugy:
> Does this: sicaria_store.modifications.advance[$sicaria_store.modifications.advance] need to use advancement as well?

Probably, since the change of the name of that tag in 1.15 or something.

white-haired-uncle:
In Crooked Ravine, if you defeat she_brother you can take her on your side. 

Then, when you return and the brothers spawn again (AFAICT you must do this many times until your units can handle being hit by two white mages in one turn), she shows up on both sides.  I was thinking about setting quests/char.she_brother_defected so she would not come back on the enemy side, but I see there's a lot of references to she_brother in upcoming scenarios.  Any thoughts?

Dugy:
>  I was thinking about setting quests/char.she_brother_defected so she would not come back on the enemy side, but I see there's a lot of references to she_brother in upcoming scenarios. Any thoughts?

Yes, that should have been a one time only event.

white-haired-uncle:
Ok, what if you decide not to have her join you?  Does she come back (like the others if you kill one or two but not all)?

white-haired-uncle:
> Does this: sicaria_store.modifications.advance[$sicaria_store.modifications.advance] need to use advancement as well?

Actually, now that I look at that it doesn't look right.  Missing ".length" on the index?

Dugy:
> Actually, now that I look at that it doesn't look right. Missing ".length" on the index?

Yeah, that part is wrong too, so it probably never worked.

> Ok, what if you decide not to have her join you? Does she come back (like the others if you kill one or two but not all)?

I suppose that bullies like them would not come again if one of them died or them turned coat.

white-haired-uncle:
In Fertile Lands, a "secondary" unit (not the actual target of the attack) was killed by shockwave, and he interogated himself.  I'm guessing in situations like this (and maybe incinerate?), unit = second_unit in the last breath event?  If so, I'm thinking just check for that and skip the dialog.

white-haired-uncle:
Getting a few "error wml: [unstore_unit]: variable 'undazzled' doesn't exist".  Looking at this, what happens if the unit is dazzled, then dies before the turn refresh?

If I'm right about this, I'd think
- simple approach would be to check if undazzled.id $targets[$t].id before unstore
- but dazzled is already an LotI thing, do we even need this event?


```
                        [for]
                            array=targets
                            variable=t
                            [do]
                                {VARIABLE targets[$t].variables.dazzled yes}
                                [unstore_unit]
                                    variable=targets[$t]
                                    find_vacant=no
                                    text=_"blinded"
                                    {COLOR_HARM}
                                [/unstore_unit]
                                [event]
                                    delayed_variable_substitution=no
                                    name=turn refresh
                                    [filter_side]
                                        side=$side_number
                                    [/filter_side]
                                    [store_unit]
                                        [filter]
                                            id=$targets[$t].id
                                        [/filter]
                                        variable=undazzled
                                    [/store_unit]
                                    {CLEAR_VARIABLE undazzled.variables.dazzled}
                                    [unstore_unit]
                                        variable=undazzled
                                        find_vacant=no
                                    [/unstore_unit]
                                    {CLEAR_VARIABLE undazzled}
                                [/event]
                            [/do]
                        [/for]

```

white-haired-uncle:
This has never worked for me, going back years.  After I kill Luke, my Destroyer still can't move in the mountains.  cant pass is empty.

- I don't know what UNREACHABLE is.  Can't find any info on said macro.
- Even if that were working, is it setting movement cost through mountains to 3 in two different ways? If so, which to keep?
- I don't see anything that cleans up after this.

```
                        [store_unit]
                            [filter]
                                side=1
                                [filter_wml]
                                    [movement_costs]
                                        mountains={UNREACHABLE}
                                    [/movement_costs]
                                [/filter_wml]
                            [/filter]
                            variable=cant_pass
                        [/store_unit]
                        [for]
                            array=cant_pass
                            [do]
                                [set_variables]
                                    name=cant_pass[$i].modifications.advancement[$cant_pass[$i].modifications.advancement.length]
                                    mode=replace
                                    [value]
                                        id=mountain_passing
                                        [effect]
                                            apply_to=movement_costs
                                            replace=true
                                            [movement_costs]
                                                mountains=3
                                            [/movement_costs]
                                        [/effect]
                                    [/value]
                                [/set_variables]
                                {VARIABLE cant_pass[$i].movement_costs.mountains 3}
                                [unstore_unit]
                                    variable=cant_pass[$i]
                                    find_vacant=no
                                [/unstore_unit]
                            [/do]
                        [/for]
                        [if]
                            [variable]
                                name=cant_pass.length
                                greater_than=0
                            [/variable]

```

white-haired-uncle:
In the final room of the final scenario, if I enter the room and leave (gryphon rider with hit and run), the AI will swarm the bottom of the room, covering the top of the stairs (45.13) and never moving off.  I tried this to no effect:

```
    [side]
        no_leader=yes
        side=2
        team_name=twisted_good
        user_team_name=_"Twisted Good"
        village_gold=2
        income=0
        {GOLD 500 650 800}
        recruit=Royal Guard,Halberdier,Iron Mauler,Master Bowman,Cavalier
        [ai]
            aggression=1
            [avoid]
                x=28,37,33,45
                y=38,21,22,13
            [/avoid]
        [/ai]
    [/side]

```

Dugy:
>  I'm guessing in situations like this (and maybe incinerate?), unit = second_unit in the last breath event? If so, I'm thinking just check for that and skip the dialog.

I did not think of this possibility. I think that simply discarding it could be a solution because there is nobody to do the interrogation, but a viable alternative would be to have some specific character do the interrogation.

> Getting a few "error wml: [unstore_unit]: variable 'undazzled' doesn't exist". Looking at this, what happens if the unit is dazzled, then dies before the turn refresh?

The main problem here is that I wanted to create _blind,_ an effect similar to _dazzle,_ but I screwed up and accidentally left a piece of _dazzle's_ code there instead of having it deal with _blind._ The part that removes _dazzle_ should have removed _blind_ instead. I think I have not discovered this while playtesting because I killed him too fast (and maybe gave him more life or resistance afterwards).

> I don't know what UNREACHABLE is. Can't find any info on said macro.

If I recall correctly, it just stands for number 99. So if a unit gets 100 movement points, it can cross what is unreachable for it and 1 additional hex.

That filter probably doesn't work because Destroyer's movement cost on mountains is undefined, which may be different from UNREACHABLE. I think the only way to get this right is to store all side 1 units and skip all units whose movement cost on mountains is less than 4 or something like that.

> Even if that were working, is it setting movement cost through mountains to 3 in two different ways? If so, which to keep?

Both. The first one is not preserved when the unit is modified in some way, the second one does not apply until it is modified in some way. 

> I don't see anything that cleans up after this.

It's meant to be permanent, if I recall correctly.

> In the final room of the final scenario, if I enter the room and leave (gryphon rider with hit and run), the AI will swarm the bottom of the room, covering the top of the stairs (45.13) and never moving off. I tried this to no effect:

I don't know why they behave that way, but I don't quite understand what your avoid filter is meant to do. I think it needs to avoid everything outside of the final zone.

spinoffs/The_Beautiful_Child/scenarios/Valley_Passage.cfg
```diff
@@ -890,50 +885,63 @@
...
+                                    [variable]
+                                        name=side1[$i].movement_costs.mountains
+                                        greater_than=0    # not null
+                                    [/variable]
+                                    [and]
+                                        [variable]
+                                            name=side1[$i].movement_costs.mountains
+                                            less_than=4
+                                        [/variable]
+                                    [/and]
```
Dugy, line 905: Units for whom it's null should pass too.

white-haired-uncle: If it's null, the unit is modified by the else block.  I need that check because null is considered less than 4.

I wonder if x=="" is equivalent to x==NULL.  But then I've seen you use this if/else format enough I assumed you were doing it for a reason (e.g. if x>3 else... is basically if x != NULL and X<=3 then ...)

I don't describe that well, but this works.  I tested it a lot.

white-haired-uncle (commented):
this was supposed to be a reply, not a review

white-haired-uncle:
The commits for "dazzle -> blind" aren't tested, but I think they complete what you described was not completed.  I don't have a good way to test them.  My feelings won't be hurt if you just skip them, what was there before threw up an error, but I don't think it broke anything.

I think I'm done with this one.  There's some stuff left in the original comment marked TODO
   - A couple I have no idea what to do with.
   - Balancing changes are definitely needed, but with as many changes as we've made here, I'd leave that for a later time.

spinoffs/The_Beautiful_Child/scenarios/Valley_Passage.cfg
Dugy: All right.

PR #626, opened 2023-07-30
Luacheck: don't warn if wesnoth.effects is modified (it's supposed to be)

edwardspec:


Dugy:
Thank you.

If I find some other warning that gives false positives for harmless things, can I use a similar change to fix that too? Will it work?

edwardspec:
Yes. By default, Luacheck considers any global variable to be bad, however `globals` and `read_globals` in `.luacheckrc` allow us to list the necessary exceptions (such as "wesnoth" and "loti").

There are some warnings ("unused variable" and alike: W212, W213, etc.) that are better ignored locally (e.g. `-- luacheck: ignore 212/item_sort` in `tutorial.lua`), because ignoring them globally may result in an actual error going unnoticed.

"Unused variable" are particularly useful warnings, because they can catch a typo in the name of variable.

Dugy:
That makes sense. Thanks for explaining that.

PR #627, opened 2023-07-30
AE 1.16

white-haired-uncle:
Add some [/unit_type]

Fix some bad for/foreach tags (my bad)

### DIFF
--- a/spinoffs/Affably_Evil/scenarios/Strych.cfg
+++ b/spinoffs/Affably_Evil/scenarios/Strych.cfg
@@ -1376,7 +1376,7 @@
                                 [/then]
                             [/if]
                         [/do]
-                    [/foreach]
+                    [/for]
                     [unstore_unit]
                         variable=krux_store
                         find_vacant=no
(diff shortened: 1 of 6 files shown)


spinoffs/Affably_Evil/scenarios/Strych.cfg
```diff
@@ -561,6 +561,7 @@
                         message=_"I will be on my way."
                     [/message]
                     {AE_TRANSITION 14-30 23-25 Southern_Estates _"Southern Estates" 24 23}
+[show_objectives][/show_objectives]
```
Dugy, line 564: Bad indentation.

white-haired-uncle: There's a bug in the install when building from source, and I've had several wesnoth versions on that box.  Who knows which wmlindent is installed.  I'll check.

Or maybe I just missed running it on one.  More likely.

white-haired-uncle:
I just noticed that in the walkthrough I mention using incarcerate in Marilyn.  So maybe it belongs after all?  It would make sense, that scenario is kind of difficult, even once you learn the trick.

white-haired-uncle:
Okay, I see it now.  I cheated (set enemy HP) to check that it was included, but that doesn't trigger setting can_incarcerate.

PR #629, opened 2023-07-30
Stop using advancements patching for multiplayer

Discontinuum:
No need to use unreliable old-way event patching for multiplayer, extra advancements can be added to mp scenarios the same way they are added in the campaign


Dugy:
True. Thanks.

PR #633, opened 2023-08-01
AE

white-haired-uncle:
Put objective note back in Marilyn.

Added show_objectives when they change, for people like me who skip a lot of dialog.

I'm not the only one who finds the first scenario very difficult.  It probably turns a lot of people off the campaign.  The enemy gold/income is too high (and they're far away) and the only strategy I've found for winning somewhat consistently (assuming decent luck in the initial drop) is not very intuitive.  And then there's the incarceration threshold.  I regularly see enemies with 5-7 HP, but almost never with 4 or less, and any that are low are going to die by my ally or run to a village.  Might help a bit (lot?) if I had poison.  Propose raising the HP threshold for incarceration on easy/normal, both to make the scenario easier and to make incarceration a bigger part of the scenario as I suspect was intended.  This also affects Marilyn, but that one is difficult enough relative to the rest of the scenarios that at worst it won't hurt.

### DIFF
--- a/spinoffs/Affably_Evil/scenarios/01_Incarcerating.cfg
+++ b/spinoffs/Affably_Evil/scenarios/01_Incarcerating.cfg
@@ -12,9 +12,9 @@
     [side]
         type=Krux_father
         id=Krux
-    name=_"Krux"
+        name=_"Krux"
         canrecruit=yes
-    unrenamable=yes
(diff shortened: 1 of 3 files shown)


Dugy:
I am fine with the suggested balance changes, it happens that the difficulty slightly drifts as things change.

PR #637, opened 2023-08-03
fix missing XP

white-haired-uncle:
In Path to Glory, I never got XP for killing the dracolich.  This fixes that case, and another I hadn't noticed.

In each round, when the last unit of side 2 is killed a die event calls enemy_check, which might store side 1 with kill.  If it does so, since the die event has not completed, the XP for the kill goes to the dead instance of the killer.  At least, I'm pretty sure that's what was happening.  So I give them the XP manually.

If a unit dies from a secondary effect, like explosive attacks, no XP is given.

XP now given if a unit dies from incinerate.

enemy_check is also called from new turn "#For the case of something went wrong".  In that case, there's no unit/second, so no one gets XP.

Tested on 1.16.9.  Tested incinerate kill, kill w/o advancement, kill with one or more advancements, and just for laughs kill from incinerate but incinerator has died.

### DIFF
--- a/scenarios8/09_Gladiatrix-The_Path_to_Glory.cfg
+++ b/scenarios8/09_Gladiatrix-The_Path_to_Glory.cfg
@@ -907,6 +907,7 @@
                             speaker=moderator
                             message= _ "Oh yes, the hellhounds were defeated by our brave heroes. Do you know what awaits them next?"
                         [/message]
+                        {HEAL_VRITRA}
                         [message]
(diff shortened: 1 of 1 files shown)


Dugy:
Yes, it's possible that storing a unit during a kill event will prevent the killer from obtaining exp.

Dugy:
It looks like you have committed merge-conflicted files in 5f409fc.

PR #641, opened 2023-08-04
Remove old chapter defines

### LINKED ISSUE #640: Old style chapter/part defines
I noticed recently that when I load a save from part 2, it says "Legend_of_the_Invincibles_VI".  I think this is a simple fix, in _main.cfg just change id=.  While we're looking at _main.cfg, is that "Legacy stuff" still needed, it looks like it's just ifdefs for things that will never be defined (maybe OLD saves?)?

But that got me looking.  It appears there are a number of references to what I assume to be the old style of dealing with chapters/parts.  I'm thinking maybe the first scenario of each chapter should undefine the setting from the previous chapter (done here because scenario01 will almost always be the first, while what is now the last scenario in a chapter may not always be) and set a new one for this chapter?  However, this would make changes (such as changing unit hitpoints).  It would be "correcting" them back to what they were supposed to be (at one time), but it's also changing the behaviour that has existed, presumedly well enough, for some time.

It looks to me like for the scenarios, these could just be replaced without the ifdef/ifndef, since we know what the outcome be every time.  For the units/hybrids_of_steel.cfg, it's not so simple, there are lots of units affected.

Personally, I'd get rid of all the ifdef/ifndefs and just assume that the way it's been working with those values never defined is now the intended behaviour.

And of course, just leave it is always an option.

```
scenarios2/02_Gods.cfg:#ifdef CAMPAIGN_LEGEND_OF_THE_INVINCIBLES_II
scenarios2/03_Forest_of_Witchery.cfg:#ifndef CAMPAIGN_LEGEND_OF_THE_INVINCIBLES_II
scenarios2/04_Darkness_Closing_In.cfg:#ifdef CAMPAIGN_LEGEND_OF_THE_INVINCIBLES_II
scenarios2/07_Infinite_Legion.cfg:#ifdef CAMPAIGN_LEGEND_OF_THE_INVINCIBLES_II
scenarios2/07_Infinite_Legion.cfg:#ifndef CAMPAIGN_LEGEND_OF_THE_INVINCIBLES_II
scenarios2/08_Sightseeing.cfg:#ifndef CAMPAIGN_LEGEND_OF_THE_INVINCIBLES_II
scenarios2/10_Lilith.cfg:#ifdef CAMPAIGN_LEGEND_OF_THE_INVINCIBLES_II
scenarios4/07_Nemesis.cfg:#ifdef CAMPAIGN_LEGEND_OF_THE_INVINCIBLES_IV
scenarios4/07_Nemesis.cfg:#ifndef CAMPAIGN_LEGEND_OF_THE_INVINCIBLES_IV
scenarios8/03_United.cfg:#ifdef CAMPAIGN_LEGEND_OF_THE_INVINCIBLES_VIII
units/hybrids_of_steel.cfg:#ifdef CAMPAIGN_LEGEND_OF_THE_INVINCIBLES_IV
units/hybrids_of_steel.cfg:#ifdef CAMPAIGN_LEGEND_OF_THE_INVINCIBLES_V
```

white-haired-uncle:
When there was a choice between retaining the current behaviour or what was originally intended behaviour the current was changed, so no effect should be seen except the name when loading saves.

Closes #640


Dugy:
The old chapter starts were removed many years ago, it's unlikely a save from that era will work without lots of other problems.

Discontinuum:
I'm a bit suspicious about changing the second campaign's id, although Part 2 saves seem working on the first glance

white-haired-uncle:
> I'm a bit suspicious about changing the second campaign's id, although Part 2 saves seem working on the first glance

They seem to, except you can't change the difficulty for a save from a campaign that was started with the old id.

I suppose it should be possible to dig into the save file and figure out why and then add a workaround that would at least allow changing difficulty for new saves made from old campaigns.

Discontinuum:
Maybe better to let the id be as it was instead of figuring out a workaround? Which problem did that id change solve?

white-haired-uncle:
The only one I know of is when you went to load a saved game it said "LotI_VI" instead of "LotI_II".

Trivial, but wrong, and should be fixed at some point and now is no worse than later, AFAICT.

Discontinuum:
> when you went to load a saved game it said "LotI_VI" instead of "LotI_II".

Where was it saying it? I didn't see it, so nothing to fix
изображение

Being unable to change difficulty is a much bigger deal imo

Discontinuum:
I guess you see the id because you run Wesnoth in the debug mode. But common players aren't supposed to run it this way, so I believe that change for them isn't worth the annoying break of saves compatibility

PR #644, opened 2023-08-06
improved remove incineration

white-haired-uncle:
Picked up this little gem from the forums.


Dugy:
Oh, I didn't know that this type of filter exists.

PR #645, opened 2023-08-06
Keep the old id to preserve compatibility with old saves

Discontinuum:
The recent change of the second campaign's id breaks the compatibility with older saves: they load but players can no longer change difficulty mid-campaign. I myself had to change difficulty when I played LotI first time, so keeping that working is necessary in my opinion. IDs aren't seen by players anyway (except when in debug mode, which is not the mod in which most players play anyway), and being able to change difficulty is more important

### DIFF
--- a/_main.cfg
+++ b/_main.cfg
@@ -122,7 +122,7 @@ Legend of the Invincibles, Part II: Into the Light"
     {LOTI_EXTRA_ADVANCEMENT_LINES}
 [/campaign]
 [campaign]
-    id="Legend_of_the_Invincibles_II"
+    id="Legend_of_the_Invincibles_VI"
     name= _ "Legend of the Invincibles"+ "
 " +_"Part II:  Into the Light"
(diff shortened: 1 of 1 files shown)


Dugy:
Back in the day when I was playing Wesnoth a lot, changing difficulty was not a thing!

The difficulty change does not work even if you load the save file, create a new one and try to load that one?

Discontinuum:
Changing difficulty is available only for start-of-scenario saves. I tried to complete one scenario, the next one still doesn't allow to change difficulty.

But I actually found a much worse issue: LotI 4-5 lvl advancements stop working in older saves if you change the campaign's id. They are defined inside the ``[campaign]`` tag https://github.com/Dugy/Legend_of_the_Invincibles/blob/c410ab4537742b847abd7b76f9858d9cfc6775e5/_main.cfg#L162 and when the id changes, Wesnoth can't find the corresponding entry and load them

Dugy:
In that case, this fix is necessary and the internal name of the second part of the campaign is impossible to change.

PR #647, opened 2023-08-13
Minor performance optimization in mpsafety:run_immediately()

edwardspec:
Replaced numerous string comparisons with a hash table lookup.


Dugy:
Good idea.

PR #648, opened 2023-09-11
Fix nil error if [damage] special lacks id

Discontinuum:
Playing in a custom era, I found some unit being unrecruitable because of a lua error in stats.lua:572. The unit had a ``[damage]`` weapon special without id which caused the code to fail (match receieved nil as the 1st arg)

### DIFF
--- a/lua/stats.lua
+++ b/lua/stats.lua
@@ -569,7 +569,7 @@ function wesnoth.update_stats(original)
 		local best_backstab_mult = 0
 		local best_backstab
 		for i = #specials,1,-1 do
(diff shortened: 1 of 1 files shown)


Dugy:
Are you sure the absence of an Id won't cause trouble elsewhere? Maybe it would be better to just add the Id to that one weapon special.

Discontinuum:
Well, I'm not a maintainer of that era. From Wesnoth documentation's  point of view, absence of id is allowed, it's not a mandatory field

Dugy:
Hopefully it won't break anything else, LotI was written with the assumption the id will always be there.

PR #652, opened 2023-11-07
fix typo

dalong76:
The area search for Uria teleport locations has a typo where a 2 (two) instead of 3 (three) makes the unit location search always fail for one option.


Dugy:
Good observation.

PR #654, opened 2024-01-17
In ch8 scenario The Desert, undead no longer spawn on keeps.

Chergnobarp:
It prevents the leader from recruiting when they do. They can still spawn on castle tiles, which reduces the amount a leader can recruit.


Dugy:
Wow, I didn't think of that.

PR #655, opened 2024-01-17
Now clearing dazzle on victory.

Chergnobarp:
This happened to me not too long ago.

1. Efraim got dazzled on turn 61, side turn 11.
2. I finished the scenario on turn 62, side turn 1.
3. Turn 62, side turn 11 never ran, so dazzle wasn't removed. It carried over to the next two scenarios before I intervened.

### DIFF
--- a/utils/abilities_events.cfg
+++ b/utils/abilities_events.cfg
@@ -6715,5 +6715,47 @@
         [/for]
         {CLEAR_VARIABLE undazzle}
     [/event]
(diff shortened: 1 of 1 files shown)


Dugy:
I copied dazzle from some examples of abilities and weapon specials and didn't properly change it to the way how similar abilities are done in LotI, and it seems to be still causing occasional issues.

PR #656, opened 2024-01-17
Fix endpoint of crafting item list

dalong76:
Post on forums explains the issue. Fix works for me. https://forums.wesnoth.org/viewtopic.php?p=685914#p685914


Dugy:
Yeah, that looks like an off by 1 error. Thank you for making the fix.

PR #657, opened 2024-01-21
Picking up gear on icy cave walls

### LINKED ISSUE #375: Bad Beelzebub placement in LotI2-Where the Flames Freeze
Due to the location of the stone, it is not possible to get all the items he drops in this board. You've done a great job in other boards to give non-picked up items that are unreachable to the user at the end of the scenario, but in this level that does not seem to work.

image

You can use the same game to reproduce the victory scenario and see no items are picked up (other than what the leader drops when you kill him).

[LotI2-Where the Flames Freeze Turn 23.gz](https://github.com/Dugy/Legend_of_the_Invincibles/files/7530260/LotI2-Where.the.Flames.Freeze.Turn.23.gz)

Thanks. Still love the campaigns!

Chergnobarp:
Fix #375 

Just started ch9, it'll be a while before I get to a scenario where I can test this.

### DIFF
--- a/utils/abilities_events.cfg
+++ b/utils/abilities_events.cfg
@@ -6715,5 +6715,47 @@
         [/for]
         {CLEAR_VARIABLE undazzle}
     [/event]
+    [event]
+        name=victory
+
+        [store_unit]
+            [filter]
+                status=dazzled
+            [/filter]
+            variable=undazzle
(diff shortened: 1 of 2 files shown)


Dugy:
I decided to be optimistic and not leave this hanging until you are sure it's working.

PR #658, opened 2024-01-21
fix corrupted elvish fighter graphic

dalong76:
fix error in path to graphic


Dugy:
Thank you.

Discontinuum:
A little problem, it won't work in 1.16. Maybe we should either make a separate branch for 1.17/18 fixes or #ifvers

dalong76:
I've also realized it doesn't totally solve the problem. The sword attack graphics have the same problem. I'll look into it some more.

PR #664, opened 2024-02-26
Update Manhunter.cfg

white-haired-uncle:
[/array] -> [/for]


Dugy:
Was it preventing the campaign from starting?

white-haired-uncle:
Nope.  Guess it would only object if that unit was invoked perhaps?

wmllint found it while looking for 1.17/18 stuff

Dugy:
I don't think that would have changed anything. If it is in the folder, it will be parsed.

PR #665, opened 2024-02-27
Update 03s_In_the_Magical_Mirror.cfg

white-haired-uncle:
1.17: The [kill] tag now explicitly no longer accepts [filter].

Scenario aborts in 1.17 if [filter] included.  Works in 1.16.10 and 1.17.25 when removed.  Not sure about anything earlier.

https://forums.wesnoth.org/viewtopic.php?p=686708#p686708

### DIFF
--- a/scenarios2/03s_In_the_Magical_Mirror.cfg
+++ b/scenarios2/03s_In_the_Magical_Mirror.cfg
@@ -719,9 +719,7 @@
         [/message]
         [kill]
             side=5
-            [filter]
-                race=elf
-            [/filter]
+            race=elf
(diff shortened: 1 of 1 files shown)


Dugy:
Thank you. Does this problem appear also elsewhere or is it only in this scenario? I think the `[filter]` tag inside `[kill]` is always a mistake.

white-haired-uncle:
Not sure.  Supposedly wmllint will find them, but I haven't run it.

Dugy:
Okay, I suppose it will be found eventually.

PR #667, opened 2024-02-27
genre -> gender

white-haired-uncle:


Dugy:
I forgot about it, sorry.

Actually, fixing legit typos is a good way to become a contributor, which grants better privileges with the repo, which can allow some exploits, so __legit fixing a typo__ pull request is not as trivial as it seems (that's not a case here since the account is already a contributor).

Dugy:
There are some privileges granted even to contributors, such as starting the CI check. And that can be exploited sometimes, because you are running your own code on the CI server.

PR #671, opened 2024-02-28
[replace_map] map= -> map_data=

white-haired-uncle:
Tested in 1.16.10 and 1.17.25.

map= still works, but tosses a warning in debug

Not sure how that extra commit snuck in there.


Dugy:
I have merged the second commit. I think you got the first commit there by making this branch by forking the branch of the other pull request.

PR #673, opened 2024-02-28
move skipped message to [story]

white-haired-uncle:
Kind of guessing what was intended here.  The message was being skipped the way it was.

Tested 1.17.25


Discontinuum:
Skipped? Strange, I think I remember reading it... But maybe I'm wrong

white-haired-uncle:
> 20240228 10:03:27 error wml: [message] not supported at scenario toplevel

PR #674, opened 2024-02-28
wesnoth.fire -> wml.fire

white-haired-uncle:
Renamed post 1.14.   Tested 1.16.10 and 1.17.25.


Dugy:
Thanks. I am surprised it's the only occurrence.

PR #675, opened 2024-03-03
Fixed Typos

Bafy78:


Dugy:
Some of these weren't typos but rather my imperfect English. Thanks for the corrections.

PR #677, opened 2024-03-04
xbow.png -> crossbow.png

white-haired-uncle:
There is no xbow.png


Dugy:
Good observation.

Dugy:
The file name is `cloak-dark.png`. I followed the naming patterns from wesnoth's core, but they are not very consistent with themselves and I've made mistakes.

PR #684, opened 2024-03-05
update dep lua functions

white-haired-uncle:
Tested on chapters 1-6.

No immediate need for this, just removes warnings.


Dugy:
Thanks.

PR #685, opened 2024-03-06
placement=map_passable -> use placement=map and passable=yes

white-haired-uncle:
> map_passable: If x,y are explicitly given and point to a valid on-map location - try to place unit at this location; if the hex is of an impassable terrain for the unit being placed, or is already occupied by another unit, the new unit will be placed in the nearest vacant hex. ([Version 1.13.8 and later only](https://wiki.wesnoth.org/DevFeature)) Deprecated, use placement=map and passable=yes instead.


Dugy:
Thanks.

PR #689, opened 2024-03-08
lesser nightstalk forest terrain

white-haired-uncle:
Gave Ef nighwalker in Jungle Hell, at night, and it didn't do anything.  It looks like maybe the terrain type is newer than when this ability was created or something?  From the description, it sounds like we should just match all forest tiles?

Worked when I tested it, but all I know about terrain is what I just looked up.

### DIFF
--- a/utils/abilities.cfg
+++ b/utils/abilities.cfg
@@ -958,7 +958,7 @@ Does not affect poisoned enemies."
         [filter_self]
             [filter_location]
                 time_of_day=chaotic
-                terrain=*^Fp,*^Fet,*^Ft,*^Fpa,*^Fd*,*^Fm*,*^V*,H*,M*,C*,K*,*^Zf*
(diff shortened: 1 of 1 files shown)


Dugy:
I am surprised that actually did anything, because `*^F*` should have matched all the terrain types you listed instead of it. Jungle Hell uses a lot of custom terrain, but they all start with `^Zf` and should have been matched by that.

Anyway, since I have no idea why your fix works, can you please put back the `*^F*` symbol to the list of terrains? I will merge it then.

white-haired-uncle:
I think you're looking at the diff backwards.  I put in `*^F*`, the list is what I replaced.

Most of the forest in JH is `Gs^Ftp` according to the gui.

Dugy:
Ah, yes, sorry, I've made a stupid mistake.

PR #692, opened 2024-03-10
confirm go to scenario

white-haired-uncle:
Looks a little nicer, allows one-click (or better yet just enter/escape) selection.


Dugy:
Good idea.

PR #694, opened 2024-03-12
Don't unstore Uria when she's dead

white-haired-uncle:
`error wml: [unstore_unit]: variable 'uria_store' doesn't contain unit data
`

Perhaps it would be better to remove the event when you kill her, but I hate testing wml and this works.

### DIFF
--- a/scenarios9/33_Pandemonium.cfg
+++ b/scenarios9/33_Pandemonium.cfg
@@ -368,6 +368,11 @@
                     name=turn refresh
                     first_time_only=no
                     id=uria immortality occasional
+                    [filter_side]
+                        [has_unit]
+                            id=Uria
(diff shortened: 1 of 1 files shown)


Dugy:
Removing events is tricky, this is the easier and more reliable way of doing it.

One way of testing WML quickly is to find the snapshotted event in the safe file and edit it there, then just reload the save file. It has some drawbacks, like no macros and no error messages on wrong syntax, but it's quite doable if the code is written and just need to debug and patch a few things. Actually, some parts of LotI contain no macros even for very usual things because the code was pasted back from a save file after a debugging session.

white-haired-uncle:
Good tip, I'll have to save that one so I don't forget it.  Thanks.

PR #696, opened 2024-03-17
Add a castle option

white-haired-uncle:
So, a lot of changes here, mostly just formatting (wmlindent doesn't seem to like right paren after closing tag). The interesting stuff happens at line 311 in the scenario diff.  And whatever the map editor did.

Fimbulwinter is still a complete PITA.  I guess you're supposed to use the runes (?), but this scenario is too hard, and comes after a terrible one -- I'm not stepping on any magic mumbo-jumbo.

You start with some gold (if you have any left from the previous), but not enough to recruit all the fodder you'll need (and even if you did, moving that much dead weight every turn sucks).  Also, your units are pretty weak at this point, so you have to wait and pick off the demons when they are in groups of no more than two, which makes the turn limit quite hard (if you get really lucky, otherwise it's just impossible).  You can't afford to leave a hero back to recruit, or race for the border, or return to recruit later, as you need them on the front lines.

Around half way, you've got plenty of income/gold, but no way to spend it, and probably a need for more fodder.  So I added a castle about where you'd need it.  Not being able to explain why it's there (which is honestly no different than about every single wesnoth scenario ever), I made it ancient to avoid the issue.

Then I added an option to spend some gold to repair it to working order if you want to use it.  I like this.  If the player finds the scenario too hard, they can buy an out (or at least a chance).  If they don't find it too hard, they can continue cheating and just skip the option and nothing changes.

The castle's a bit off the main path, so there's a "hey look" message on easy.


Dugy:
The diffs are broken because indentation was changed in much of the file. It was triggered by the correction you spoke of, right?

I am not 100% certain how is that keep supposed to fix the balance problem you're describing. Do you mean that the scenario start is still hard, but it gets easier over time as the player captures villages and can make use of the gold he starts to gain?

If I recall correctly, in that scenario, you could capture villages among the wandering demons and then recruit more to secure the area, but you found it inconvenient to recruit from the starting keep?

white-haired-uncle:
> The diffs are broken because indentation was changed in much of the file. It was triggered by the correction you spoke of, right?

Yes.  If you look at the original (if statement starting at line 245, pretty much everything starting at 514), it's off and wmlindent just makes it worse. 
> 
> I am not 100% certain how is that keep supposed to fix the balance problem you're describing. Do you mean that the scenario start is still hard, but it gets easier over time as the player captures villages and can make use of the gold he starts to gain?

Yes, still hard, but about 60% of the way through a castle will allow you to recruit some more fodder to either continue the fight or make a break for the border.  At this point, I usually have over 600 gold, I'm more or less out of fodder, and it's questionable whether I'll reach the imps before turns run out, let alone fight my way through/around.

> 
> If I recall correctly, in that scenario, you could capture villages among the wandering demons and then recruit more to secure the area, but you found it inconvenient to recruit from the starting keep?

To recruit from the starting keep, you need to keep a leader back off the front line.  At this point, only your leaders, prophet, death lord, and maybe shock trooper can really put down a powerful demon.  Since you can't avoid encountering demons in pairs, any given demon will often be immune to many/most of your attacks, you're going to miss sometimes, and you can't afford to leave any demons alive to retaliate, you really can't leave a leader back dedicated to recruiting or try to run back to the castle once you have more gold.  Also, you don't have anywhere enough turns to hang/run back.  And even if you did, moving a ton of fodder every turn is tedious.

Obviously, just slashing the number of demons would be an easy option, but getting the number right could be tricky.

Dugy:
All right.

Adding elements into the scenario is usually better than just adjusting some numbers.

Just for reference, is the player informed about the keep early enough? Or does it take a leap of faith to rush forward and hope to find something useful?

white-haired-uncle:
I don't worry about it being too far west (you don't need to rush to find it, if you don't find it by the time you need it you're probably moving too slowly anyway).  It might be too far south.  I would always find it, as I keep my army spread out to try to keep the demons spread thin, but I added an event on easy that should make sure those players don't miss it.

PR #697, opened 2024-03-17
a little help for the orcs

white-haired-uncle:
I've recently seen this described as the hardest scenario of LotI.  If it isn't, it's close.  The problem is you regularly see the orcs send a dozen or more units in a suicide charge against a single demon, who chuckles softly and levels up.  After this happens a couple times, the field is clear and the demons charge the orc leader.

At first I just wanted to paint in some hills to give the orcs a bit of an advantage, but it turns out demons get pretty much the same defenses as orcs.  So I added an icy river right about on the demon side of where the armies meet.  This worked okay, except the orcs still would charge out onto the ice, so I gave them an order to avoid ice and water, which worked quite nicely.  It's still hard, and once the demons make it across the river the orcs are no match for them, but at least now the orcs don't tend to just suicide and lose the scenario for you more or less immediately.

The concept seems solid.  I have no design skills or interest so I don't know how the map looks or if it makes sense having the river there, etc (actually the first three scenarios of ch10 always struck me as odd, travelling across a desert in hell, then green/snow, then winterland in such a short span).


Dugy:
You have correctly guessed that I have followed some basic geology when designing outdoor scenarios (I am quite fond of hiking). But parts of chapter 10 intentionally violate this rule to create an impression that everything has gone wrong with the land (the snow itself is handwaved as the mythical Fimbulwinter).

PR #701, opened 2024-03-20
agression -> aggression

white-haired-uncle:


Dugy:
How do you do it that you keep spotting all those grammar errors?

Issue #703, opened 2024-03-20
Invalid color in range: cyan


white-haired-uncle:
No idea here. 

```
20240320 00:41:48 error display: caught config::error while processing color-range RC: Invalid color in range: cyan
20240320 00:41:48 error display: bailing out from RC

spartan$ grep -r cyan
utils/item_list.cfg:        image="items/boots.png~RC(magenta>cyan)"
utils/item_list.cfg:        image=items/boots.png~RC(magenta>cyan)~CS(50,50,150)
utils/item_list.cfg:        image="items/boots.png~RC(magenta>cyan)"

```

Dugy:
What happens if you replace `cyan` with `blue`?

white-haired-uncle:
error goes away

Dugy:
I think that the problem is that `cyan` isn't defined as a colour palette. I don't know there they are defined, I am sure that `blue` is defined, maybe `teal` is also defined, and it's quite close to `cyan`.

white-haired-uncle:
teal works too.  I think it looks a little better.

cyan is used for flying, iced, and eidolon's boots.  Do you want to use same for all, or blue for one to distinguish it from the others?

Dugy:
Maybe `teal` for iced boots and `blue` for the others, or something else if you find that suitable.

PR #705, opened 2024-03-21
cyan -> blues

white-haired-uncle:
Okay, three different supported shades of blue.  I don't know if they're different enough to tell, but they don't error.

I couldn't find a list of supported colors, except for team colors.

Ref #703


Dugy:
Wow, I didn't even know that `lightblue` was an accepted colour.

Discontinuum:
I think available colours defined here https://github.com/wesnoth/wesnoth/blob/master/data/core/team-colors.cfg

Dugy:
Thanks. Seems like we have all the variants of blue now.

PR #707, opened 2024-03-22
Immune to poison

### LINKED ISSUE #663: immune to drains/poison/etc may need cleanup
User reports when a drains attack occurs against defender wearing Amulet of SomethingOrOther, defender is left with immune to poison (with loti3.2.7c):
https://forums.wesnoth.org/viewtopic.php?p=686113#p686113

Looks to me like in utils/abilities_events.cfg the attack event immunizes the defender and the attack_end removes that from the attacker (I realize that there are two pairs of events, one for attacker has immune and one for defender has immune, but the latter looks inconsistent and this matches the behaviour).

But while looking at this, I noticed a couple things:

1.  Does this approach of setting not_living for the duration of the attack make sense?  What if an attacker used poison and drain (for example) on the same attack, and the defender only had immune to drain?  
2. In 3.2.7c, is there a reason for using modify_unit to set immunity and unstore_unit to remove?  Don't suppose it really matters, but consistency and all.
3. In 3.2.7c, the "wrong unit issue" applies to both poison and drains
4. In 3.2.7c, there are three pairs of immune to poison attack events (I'd expect two).  It looks like maybe the extra one was an attempt to fix the "wrong unit issue" but was implemented as an addition instead of replacement?
5. In beta, I don't see any reference to immune to drain.  EDIT:  Okay, now uses undrainable.  Maybe much of the above has been addressed, but there's just some cleanup left to do?

white-haired-uncle:
Tested with Intoxicator, and Scourge of All Living, v1.19.0-dev (94b71ac95b2-Clean)

Closes #663


Dugy:
Thanks.

white-haired-uncle:
Forgot to mention something.  Irongut Helmet checks for not_living before applying immune to poison.  It's the only place that does that.  I can't think of any repercussion for doing so, or not, but it's inconsistent with other items/amlas/whatever, and I don't like the way it is right now (potentially set unpoisonable -- on undead -- but don't set 'immune to poison').

The only implication I could think of is an undead unit losing "undead" should still have immune to poison if holding Intoxicator, for example.  Could this happen in MP with Purify?

P.S.  I've always wanted to try Purify, but it's a late Leg of Light AMLA and by the time it's available I'm usually down to fighting mechanical/demons but no undead.

```
        [effect]
            apply_to=new_ability
            [filter]
                [not]
                    status=not_living
                [/not]
            [/filter]
            [abilities]
                {ABILITY_IMMUNE_TO_POISON}
            [/abilities]
        [/effect]
        [effect]
            apply_to=status
            add=unpoisonable
        [/effect]

```

Dugy:
I think that check for being undead is unnecessary now. The effect is enough. The ability might be useful to make it visible.

There shouldn't be a problem with the Purify ability, because it turns enemy undead killed into living units, but enemy undead won't be wearing this item. Also, I think the filter would be checked again after the transformation.

Dugy:
There are only coop scenarios. I think someone could hack some PvP, but if I remember correctly, inventories are still shared between all humans in the game, so it's not really something supported.

But if it works the way I think it does, the filter will be rechecked if the unit transforms to another type, so this item will work as expected.

Issue #691, opened 2024-03-09
frail tide / conviction not included in unit info penetrates table


white-haired-uncle:
In Unit Information, the description for frail tide and conviction are shown (and the abilities do work) but the resistance/penetrates table does not include the penetrates.  It looks like unit info looks for resistances with id=fire_penetrates (for example), which is what most items/amlas seem to set.  

I'm wondering if frail/conviction are the only things that affect multiple (physical, non-physical) resistances.  Only ones I can think of.  EDIT: nope, dark dragon legacy.

Also, it looks like the report just uses the last [resistance] sub it finds for an attack type.  What if there are more than one (could there be)?  How does bfw handle that during combat?  [Note to self: And is it always the same, or is this finally a clue to why I can get such great cold penetration with my CMs???]

white-haired-uncle:
Turn the effect on/off at the beginning/end of an attack?

Dugy:
No need, I have figured out a better way of doing it, it's even included in the damage calculation.

Dugy:
Okay, I have changed how penetrate works in commit 5dcc500.

white-haired-uncle:
That sort of works.

Efraim's cold penetration from legacy still affects Lethalia's blizzard attack (and the target stats) on the flies.

The Gryphon Rider's arcane penetration from Void armour still affects Lethalia's fairie fire (and the target stats) on the flies, unless you drop the item to the ground and pick it up in which case it seems to work as intended.

I'm guessing it would work for specials earned by legacy and other means, but only for new instances?

I think I like this, but I also think it's still confusing.  Even if a new field was added to the res/pen table to reflect ability based "penetration", still need to explain to the user why they are different in a clear way.

[LotI2-Infernal Narrows-Auto-Save3.gz](https://github.com/Dugy/Legend_of_the_Invincibles/files/14546889/LotI2-Infernal.Narrows-Auto-Save3.gz)

Dugy:
Yes, you need to cause the unit's properties to be updated somehow. Changing something in the inventory is the easiest way of doing that on purpose (you can change it back right afterwards). If this wasn't a lua-only change, you would have to start a new scenario and update the properties later, but it's not necessary in this case.

white-haired-uncle:
[...]

PR #708, opened 2024-03-22
Frail tide/conviction

white-haired-uncle:
Ignore (better yet nuke) the "Merge pull request 1", I was experimenting with my fork (wanted to learn how to apply individual commits instead of just syncing with yours) and now it's forked.

Three commits here to pick and choose from:

1. Added a book to Hunter/Legacy to make getting ft/convict together more difficult.  I did not try to figure out the drop weights, they are whatever I copied them from.  [I was toying with the idea of a persistent variable set in AE/The_Reckoning when you pick up Agrippa's book -- places it in your inventory at the start of LotI (chapter10 I think).  No real reason, just a bit of a fun tie-in no one would ever notice]
2. Make prophet/CM take other AMLAs before ft/convict.  The way I advance them this may be little more than an incovenience.  Baby steps.
3. Weaken CM conviction from 15/25/33 to 10/15/20.

Tested v1.19.0-dev (94b71ac95b2-Clean)

Ref #691 

BTW, while testing I noticed that when you pick up a set item, it doesn't show other "requires" items in bold like it does in Items.  Did it not used to do so?  Not that I can really tell (purple on black???), but I thought it used to work that way, and/or it would be useful.


Dugy:
Looks okay. Except for one thing. The book is quite wrong. First, is it supposed to give 10% to arcane resistance or 15% arcane resistance? Second, that item set was not supposed to refer to Agrippa. I will not explain the riddle, but I will elaborate on it when I tell you that the book should be named _Lost Memoirs of the Inglorious_ and the flavour text should be _Nobody knew his identity, yet everyone knew his face._

PR #714, opened 2024-03-26
canrecrut -> canrecruit

white-haired-uncle:
Please also note the comments on gold/income for our side being higher on harder levels, particularly Chasing Dragons which is just ridiculously easy.

### DIFF
--- a/scenarios2/01_The_Beginning.cfg
+++ b/scenarios2/01_The_Beginning.cfg
@@ -21,7 +21,7 @@
         side=1
         controller=human
         recruit=Elvish Fighter,Elvish Archer,Elvish Shaman,Elvish Scout,Elvish Druid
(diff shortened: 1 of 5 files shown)


Dugy:
Uh, yeah, it looks like the enemy gold increase with difficulty isn't very steep. I could increase it, or add some extra enemy there for higher difficulties.

white-haired-uncle:
I just find it odd to increase allied gold/income on harder levels.  Seems backwards if anything.  If it's intentional, it should probably be commented, because to me it looks like a typo.

Dugy:
I think it's useful for allies because it improves their survivability if they start somewhere away from you. Even if their survival is not required, they still add a bit of uniqueness to the scenario.

PR #650, opened 2023-10-06
Move halos to abilities(wesnoth 1.18 and later)

newfrenchy83:
after https://github.com/wesnoth/wesnoth/commit/a17369597c4869938d0a44a70bd9708452c0fd6b, it is possible what abilities
support directly halos image_halo atttribute is usded for give halo to units under influence of abilities when active,
halo_image_self give halo to owner of abilites when active(filter and attribute specific of abilities used like specials wasn't concerned)

PR #715, opened 2024-03-26
Add level Harder

white-haired-uncle:
The existing difficulty levels just aren't (aside from a handful of scenarios).  I added a fourth level to experiment with adding some amount of difficulty.  I was just targetting "Normal", but I think I actually came up with something approaching Hard.  It's kind of a tricky question, the scenarios are certainly a bit harder but I've yet to see the longer term effects.  For example, I usually finish ch1 with about 500 saved_turns, testing on Harder I've ended up close to 300.  Also, I've actually lost experience/geared units (though not enough) -- my goal if I see this through is not to end each part with 300-600 unused items and no losses (except Exec Sanct, where I just let senior units die to finish as fast as possible).  So it's tough to say how these effects will play out in chapters to come.

I've played this through a couple times, and most scenarios several times each (all on 1.19) and while it's not quite as hard as I'd like I think it's close enough to put it out there for feedback and to see if there's any interest, both on your part and the players'.  If you're interested, I'm wondering if it's possible (I assume it must be), to create a branch/fork and publish a separate experimental version. 

I tried not to touch the existing levels at all, but there were a couple places where it was just easier to change ifdef HARD to ifndef EASY, for example, and I only did that on scenarios that were really trivial anyway.

BTW, I've never relied much on elves (save gryphon riders), but on my last run I have a Sylph and an Enchantress who are nearly as powerful as Lethalia (at least, they almost always kill any enemy unit in one turn) even with minimal gear.  And Lethalia seems MUCH more powerful than normal.  It's all fairie fire.  I suspect the recent changes to the elves are pretty major.

### DIFF
--- a/_main.cfg
+++ b/_main.cfg
@@ -108,6 +108,12 @@
         label=_"Tyrant"
         description=_"Hard"
     [/difficulty]
+    [difficulty]
+        define=HARDER
+        image=units/human-loyalists/shocktrooper.png~RC(magenta>red)
+        label=_"Trooper"
+        description=_"Harder (Experimental)"
+    [/difficulty]
     description= _ "Sometimes, people don't embrace the tempting powers of darkness because of desires to rule, but merely for survival. Can the dark powers be used to fight against dark schemes with no limits or will they eventually corrupt its users?
 
 (Intermediate level, 90 scenarios separated to five chapters)
--- a/lua/main.lua
+++ b/lua/main.lua
@@ -194,7 +194,7 @@ function wesnoth.wml_actions.harm_unit_loti(cfg)
 						       )
 
 			if unit_to_harm.hitpoints <= damage then
-				if kill then 
+				if kill then
 					damage = unit_to_harm.hitpoints
 				else
 					damage = unit_to_harm.hitpoints - 1
--- a/lua/main.lua
+++ b/lua/main.lua
@@ -3,8 +3,6 @@
 loti = {}
 loti.util = {}
 
---fake change
-
 -- Syntax sugar to make GUI2 dialogs more compact, e.g. T.column, T.row, etc.
 T = wml.tag
 
--- a/lua/main.lua
+++ b/lua/main.lua
@@ -136,6 +136,8 @@ function wesnoth.wml_actions.harm_unit_loti(cfg)
 			local animate = cfg.animate -- attacker and defender are special values
 			local delay = cfg.delay or 500
 			local fire_event = cfg.fire_event
+			local kill = true
+			if cfg.kill ~= nil then kill = cfg.kill end
 			local primary_attack = wml.get_child(cfg, "primary_attack")
 			local secondary_attack = wml.get_child(cfg, "secondary_attack")
 			local harmer_filter = wml.get_child(cfg, "filter_second")
@@ -192,7 +194,11 @@ function wesnoth.wml_actions.harm_unit_loti(cfg)
 						       )
 
 			if unit_to_harm.hitpoints <= damage then
-				damage = unit_to_harm.hitpoints
+				if kill then 
+					damage = unit_to_harm.hitpoints
+				else
+					damage = unit_to_harm.hitpoints - 1
+				end
 			end
 
 			unit_to_harm.hitpoints = unit_to_harm.hitpoints - damage
--- a/lua/main.lua
+++ b/lua/main.lua
@@ -3,6 +3,8 @@
 loti = {}
 loti.util = {}
 
+--fake change
+
 -- Syntax sugar to make GUI2 dialogs more compact, e.g. T.column, T.row, etc.
 T = wml.tag
 
--- a/scenarios1/01_An_Orcish_Assault.cfg
+++ b/scenarios1/01_An_Orcish_Assault.cfg
@@ -6,7 +6,7 @@
     next_scenario="02_The_Assassination"
     experience_modifier=125
     {GLOBAL_EVENTS}
-    {TURNS 16 15 14}
+    {LOTI_TURNS 16 15 14 13}
     {SCENARIO_MUSIC "elvish-theme.ogg"}
 
     {DEFAULT_SCHEDULE}
@@ -19,7 +19,7 @@
         profile=portraits/Efraim.png
         canrecruit=yes
         controller=human
-        gold=150
+        {LOTI_GOLD 150 150 150 120}
         recruit=Fencer,Thief,Spearman,Bowman,Heavy Infantryman,Poacher
         team_name=Loyalists
         user_team_name=_"Loyalists"
@@ -113,8 +113,8 @@
         type=Orcish Warrior
         canrecruit=yes
         recruit=Orcish Grunt,Wolf Rider,Orcish Archer,Troll Whelp
-        {GOLD 175 200 225}
-        {INCOME 1 2 2}
+        {LOTI_GOLD 175 200 225 250}
+        {LOTI_INCOME 1 2 2 3}
         random_traits=yes
         team_name=evil
         user_team_name=_"Evil"
--- a/scenarios1/02_The_Assassination.cfg
+++ b/scenarios1/02_The_Assassination.cfg
@@ -6,7 +6,7 @@
     name=_ "The Assassination"
     map_data="{~add-ons/Legend_of_the_Invincibles/maps/02_Redain_Castle.map}"
     {GLOBAL_EVENTS}
-    {TURNS 20 17 15}
+    {LOTI_TURNS 20 17 15 13}
     victory_when_enemies_defeated=no
 
     {SCENARIO_MUSIC loyalists.ogg}
@@ -32,7 +32,7 @@
         unrenamable=yes
         canrecruit=yes
         recruit=Fencer,Thief
-        {GOLD 60 50 45}
+        {LOTI_GOLD 60 50 45 20}
         income=0
         shroud=yes
         fog=yes
@@ -144,6 +144,22 @@
             canrecruit=no
             random_traits=yes
         [/unit]
+#ifdef HARDER
+        [unit]
+            type=Royal Guard
+            side=2
+            x,y=3,6
+            canrecruit=no
+            random_traits=yes
+        [/unit]
+        [unit]
+            type=Royal Guard
+            side=2
+            x,y=1,6
+            canrecruit=no
+            random_traits=yes
+        [/unit]
+#endif
 #ifdef HARD
         [unit]
             type=Swordsman
@@ -152,8 +168,6 @@
             canrecruit=no
             random_traits=yes
         [/unit]
-#endif
-#ifdef HARD
         [unit]
             type=Swordsman
             side=2
--- a/scenarios1/03_Banished.cfg
+++ b/scenarios1/03_Banished.cfg
@@ -6,7 +6,7 @@
     {GLOBAL_EVENTS}
     {SCENARIO_MUSIC "wanderer.ogg"}
     {EXTRA_SCENARIO_MUSIC "nunc_dimittis.ogg"}
-    {TURNS 35 33 31}
+    {LOTI_TURNS 35 33 31 29}
     victory_when_enemies_defeated=no
 
     {AFTERNOON}
@@ -102,6 +102,9 @@
 #ifdef HARD
         recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll Whelp,Troll, Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior
 #endif
+#ifdef HARDER
+        recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll, Goblin Knight,Orcish Slayer,Orcish Warrior
+#endif
 
         gold=0
         income=−2
@@ -112,7 +115,7 @@
                 [effect]
                     apply_to=new_ability
                     [abilities]
-#ifndef HARD
+#ifdef EASY
                         {ABILITY_WEAK_REFLECT}
 #else
                         {ABILITY_REFLECT}
@@ -167,8 +170,8 @@
         [/filter]
         [modify_side]
             side=2
-            {GOLD 200 250 300}
-            income=15
+            {LOTI_GOLD 200 250 300 350}
+            {LOTI_INCOME 15 15 15 20}
         [/modify_side]
         [unit]
             type=Orcish Grunt
--- a/scenarios1/04_Paradise_Lost.cfg
+++ b/scenarios1/04_Paradise_Lost.cfg
@@ -60,7 +60,7 @@
         recruit=Fencer,Thief
         team_name=defenders
         user_team_name=_"Defenders"
-        gold=200
+        {LOTI_GOLD 200 200 200 175}
         controller=human
     [/side]
     [side]
@@ -93,9 +93,12 @@
 #ifdef HARD
         recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll Whelp,Troll, Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Troll Rocklobber
 #endif
+#ifdef HARDER
+        recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll, Goblin Knight,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Troll Rocklobber
+#endif
 
-        {GOLD 180 250 330}
-        {INCOME 15 22 30}
+        {LOTI_GOLD 180 250 330 350}
+        {LOTI_INCOME 15 22 30 32}
         team_name=evil
         user_team_name=_"Evil"
         [modifications]
@@ -103,10 +106,10 @@
                 [effect]
                     apply_to=new_ability
                     [abilities]
-#ifdef HARD
-                        {ABILITY_GREATER_EVISCERATION}
-#else
+#ifdef EASY
                         {ABILITY_EVISCERATION}
+#else
+                        {ABILITY_GREATER_EVISCERATION}
 #endif
                     [/abilities]
                 [/effect]
@@ -134,9 +137,12 @@
 #ifdef HARD
         recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll Whelp,Troll, Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Troll Rocklobber,Goblin Impaler
 #endif
+#ifdef HARDER
+        recruit=Orcish Grunt,Orcish Archer,Troll Whelp,Troll,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Troll Rocklobber
+#endif
 
-        {GOLD 70 150 200}
-        {INCOME 20 25 30}
+        {LOTI_GOLD 70 150 200 200}
+        {LOTI_INCOME 20 25 30 32}
         team_name=evil
         user_team_name=_"Evil"
         [modifications]
@@ -152,7 +158,8 @@
             [/advancement]
         [/modifications]
     [/side]
-    [side]
+    [side]  # No harder on HARDER, since side 1 can't get there to help in time.  May need to lower starting gold (& raise income?) anyway.
+            # May make sense to make Lethalia hard to hit for this scenario so player doesn't lose due to unavoidable circumstances.
         id=Vrongromor
         name= _ "Vrongromor"
         type=Orcish Warlord
@@ -169,9 +176,12 @@
 #ifdef HARD
         recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll Whelp,Troll, Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Troll Rocklobber,Direwolf Rider
 #endif
+#ifdef HARDER
+        recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll Whelp,Troll, Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Troll Rocklobber,Direwolf Rider
+#endif
 
-        {GOLD 140 220 300}
-        {INCOME 10 15 20}
+        {LOTI_GOLD 140 220 300 300}
+        {LOTI_INCOME 10 15 20 20}
         team_name=evil
         user_team_name=_"Evil"
         [modifications]
--- a/scenarios1/05_Shatter_the_Defilers.cfg
+++ b/scenarios1/05_Shatter_the_Defilers.cfg
@@ -7,7 +7,7 @@
     {SCENARIO_MUSIC "elvish-theme.ogg"}
     {EXTRA_SCENARIO_MUSIC "elf-land.ogg"}
     {EXTRA_SCENARIO_MUSIC "vengeful.ogg"}
-    {TURNS 40 37 35}
+    {LOTI_TURNS 40 37 35 33}
     victory_when_enemies_defeated=no
 
     {DEFAULT_SCHEDULE}
@@ -83,6 +83,7 @@
             y=5
             side=1
         [/unit]
+#ifndef HARDER
         [unit]
             type=Elvish Shaman
             generate_name=yes
@@ -97,6 +98,7 @@
             y=8
             side=1
         [/unit]
+#endif
         [unit]
             type=Elvish Marksman
             generate_name=yes
@@ -184,9 +186,12 @@
 #ifdef HARD
         recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll Whelp,Troll, Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Troll Rocklobber,Orcish Warrior
 #endif
+#ifdef HARDER
+        recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll, Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Troll Rocklobber,Orcish Warrior
+#endif
 
-        {GOLD 150 200 300}
-        income=2
+        {LOTI_GOLD 150 200 300 350}
+        {LOTI_INCOME 2 2 2 8}
         team_name=evil
         user_team_name=_"Evil"
 #ifndef EASY
@@ -221,8 +226,12 @@
 #ifdef HARD
         recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll Whelp,Troll, Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Troll Rocklobber,Goblin Impaler,Goblin Rouser
 #endif
+#ifdef HARDER
+        recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll Whelp,Troll, Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Troll Rocklobber
+#endif
 
-        {GOLD 150 200 300}
+        {LOTI_GOLD 150 200 300 350}
+        {LOTI_INCOME 2 2 2 8}
         income=2
         team_name=evil
         user_team_name=_"Evil"
@@ -251,14 +260,17 @@
         recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll Whelp,Orcish Assassin,Orcish Crossbowman,Troll Shaman
 #endif
 #ifdef NORMAL
-        recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll Whelp,Troll, Goblin Knight,Orcish Assassin,Orcish Crossbowman,Troll Shaman
+        recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll Whelp,Troll,Goblin Knight,Orcish Assassin,Orcish Crossbowman,Troll Shaman
 #endif
 #ifdef HARD
-        recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll Whelp,Troll, Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Troll Rocklobber,Direwolf Rider
+        recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll Whelp,Troll,Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Troll Rocklobber,Direwolf Rider
+#endif
+#ifdef HARDER
+        recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll Whelp,Troll,Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Troll Rocklobber,Direwolf Rider
 #endif
 
-        {GOLD 150 200 300}
-        income=2
+        {LOTI_GOLD 150 200 300 350}
+        {LOTI_INCOME 2 2 2 8}
         team_name=evil
         user_team_name=_"Evil"
         [modifications]
@@ -266,10 +278,10 @@
                 [effect]
                     apply_to=new_ability
                     [abilities]
-#ifdef HARD
-                        {ABILITY_ABSORB_4}
-#else
+#ifdef EASY
                         {ABILITY_ABSORB_2}
+#else
+                        {ABILITY_ABSORB_4}
 #endif
                     [/abilities]
                 [/effect]
--- a/scenarios1/06_The_Ruins_of_Lost_Empires.cfg
+++ b/scenarios1/06_The_Ruins_of_Lost_Empires.cfg
@@ -11,7 +11,7 @@
     victory_when_enemies_defeated=no
 
     {DEFAULT_SCHEDULE}
-    turns=100
+    {LOTI_TURNS 100 100 100 60}
     next_scenario=07_The_Return
 
     [event]
@@ -65,6 +65,9 @@
 #endif
 #ifdef HARD
         recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Troll Whelp, Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Troll Rocklobber,Orcish Warrior
+#endif
+#ifdef HARDER
+        recruit=Orcish Grunt,Orcish Archer,Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Troll Rocklobber,Orcish Warrior
 #endif
 
         gold=0
@@ -131,8 +134,11 @@
 #ifdef HARD
         recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Goblin Impaler,Goblin Rouser
 #endif
+#ifdef HARDER
+        recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman
+#endif
         ai_special=guardian
-        {GOLD 0 0 0}
+        {LOTI_GOLD 0 0 0 0}
         income=−2
         team_name=evil
         user_team_name=_"Evil"
@@ -189,7 +195,7 @@
         profile="portraits/orcs/ruler-2.png"
         random_traits=yes
         side=6
-        {GOLD 0 0 0}
+        {LOTI_GOLD 0 0 0 0}
         income=−2
         ai_special=guardian
         canrecruit=yes
@@ -203,6 +209,9 @@
 #ifdef HARD
         recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Direwolf Rider
 #endif
+#ifdef HARDER
+        recruit=Orcish Grunt,Orcish Archer,Wolf Rider,Goblin Knight,Orcish Assassin,Orcish Slayer,Orcish Warrior,Orcish Crossbowman,Direwolf Rider
+#endif
         team_name=evil
         user_team_name=_"Evil"
         [modifications]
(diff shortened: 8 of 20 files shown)


Dugy:
If I recall correctly, a 4 difficulty setup is quite expected, so you can use the `GOLD4` macro instead of defining a `LOTI_GOLD` macro etc.

I usually felt like 3 difficulties are enough, if hard is not hard enough, it can be made harder. More enemy gold is not all the difficulty that higher difficulties bring. It also makes items weaker, increases exp requirements, makes custom enemy units stronger and a few other tweaks. That is a lot of work. And also, some scenarios are reportedly significantly harder than others. That is why I always preferred making easy scenarios harder rather than thinking about adding a new difficulty.

Of course, if you want to create and maintain a fork, maybe it will attract enough players to justify the maintenance and I will merge it.

white-haired-uncle:
Ultimately, if this little experiment is successful, I'd prefer to see Harder back ported to be the new Hard (be happy to do that), leaving the standard three levels with a fourth "reserved" for Nightmare, should anyone decide to go that route (I have some ideas, but no interest in actually playing what I would consider NM).

But because it's such an undertaking, I figured the best thing would be to keep my changes completely separate for now until they can be evaluated within the scope of an entire five chapters.

FWIW, I actually suspect that increasing enemy gold is probably the least preferable approach to increasing difficulty.  If you don't first make his units harder to kill and yours easier to die, more enemy gold probably just means more XP for your team, making subsequent scenarios easier, not harder.  And it's not nearly as interesting.

I thought I remembered a GOLD4, but a search didn't find anything.

Dugy:
I am surprised that GOLD4 can't be found anymore.

> I'd prefer to see Harder back ported to be the new Hard
Then it might make more sense for you to edit the hard difficulty instead of going through all the complications of adding a 4th difficulty.

Discontinuum:
There is GOLD4 https://github.com/wesnoth/wesnoth/blob/1.18/data/core/macros/utils.cfg

But it requires the fourth difficulty to be ``NIGHTMARE``, not ``HARDER``

Discontinuum:
> Not seeing anything that requires NIGHTMARE in that.

It uses the fourth value with ``#ifdef NIGHTMARE``

https://github.com/wesnoth/wesnoth/blob/72a102ccd427caede35038bed4cd6ebebc23b29a/data/core/macros/utils.cfg#L35

https://github.com/wesnoth/wesnoth/blob/72a102ccd427caede35038bed4cd6ebebc23b29a/data/core/macros/utils.cfg#L78

white-haired-uncle:
Modified whirlwind so that it doesn't poison, slow, or kill secondary targets on Hard.  Still very useful for setting up easy kills, but unable to make huge holes to advance through.  I am not going to like this.

Dugy:
Items are already made weaker on hard, and that will bleed into other add-ons. The value on hard is the default value for situations where the difficulty defines are absent. Adding more stacks of lethargy after using AoE attacks could be a penalty in the same vein.

I don't think it's good to make abilities follow different logic on hard, though.

Dugy:
The balance problem with whirlwind is not easy to solve. The only thing that worked was to remove the exp from killing secondary targets, which resulted in players never using it. And even that didn't stop players from attempting to do easier scenarios only with leaders, because they could kill most enemies in retaliation.

The current penalty that comes with whirlwind makes retaliatory damage weaker, which takes places before any penalty falloff, so it directly counteracts the other way how leaders can cut through enemy legions. But this is not so apparent.

Dugy:
Self-conscious undead that look like humans could be subject to the _temptation_ ability, that could make sense.

But giving _temptation_ to random drakes is a rather _let's generate a bunch of random enemies with random abilities_ kind of faction. Now, _temptation_ is available only in one unit type, for whom it's their signature ability. It would just make no sense on monstrous beings like drakes.

What would be good at countering AoE effects would be damage reflection auras. Currently, thorns don't work against AoE hits and there is no thorns aura, but it can be changed. Put some 2 or 3 champions with damage reflection auras into each faction that is prone to being obliterated by a leaders AoE and the leader will get so many small hits from individual units reflecting damage that they won't be able to continue.

white-haired-uncle:
Hmm, I swear I've run into temptation in the strangest places, but the only one I can locate ATM is the bandit leader in Gods.

Using reflection is an intriguing idea, though I'm not sure what you mean by 'aura' (I've seen some things listed as aura, but never even thought about what makes X an aura vs a ???).  Clearly you've got an idea here I'm not seeing.  I assume you mean things like reflect/thorns/etc would work on AoE secondary strikes (which they don't now because there's no attacker_hits in play I assume), but I'm not sure how making an "aura" would change this.  I can think of ways to implement reflect-when-secondary, but I think you have something in mind.

Dugy:
Sorry, I was used to the term _aura_ from other games and forgot it's not well defined here. What I had in mind is an ability that affects adjacent units, in this case, I meant that the ability would make all adjacent units deal some damage back to whoever who hits them. This would be a strong deterrent to using AoE attacks against them, because taking 4 damage when hitting an enemy isn't much, but hitting 5 units affected by this ability simultaneously would result in the attacker taking 20 damage per hit, maybe 100 total damage per round, which can't be sustained.

Implementing this is nontrivial, I can do that if you want.

white-haired-uncle:
Yes, please.  It doesn't sound terribly difficult, so I obviously haven't tried.  Be interested to see how you do it.

Dugy:
I have made AoE attacks count also as attacks against other targets, so AoE attacks will trigger some on hit effects. I have also made thorns stronger, working also against non-melee attacks and added an ability to grant thorns to adjacent allies. It is available through an item, but it will be more influential if an enemy champion uses it.

The commit hash is 90e9bb5.

white-haired-uncle:
Ok, looked at thorns_aura.  If I understand you correctly, each (enemy) race would have a couple unit types which (can?) have that ability (probably their highest level units)?

So Corrupted Elvish Nightprowler and Corrupted Elvish Juggernaut might have (a chance to have?) thorns_aura?  I'm guessing you don't give them to Elves, since players can recruit those.

I assume drakes would have a couple that return fire (what's that ability like thorns with fire?), but I don't see any loti-drakes so I'm not quite sure how that would work.

Etc.

Close?

Dugy:
> If I understand you correctly, each (enemy) race would have a couple unit types which (can?) have that ability (probably their highest level units)?

No. Except for demon traits, as demons are meant to be difficult by having randomised abilities.

What I had in mind was that if some scenario is too easy to be done with leaders only, the side could start with a couple of extra units with this ability. This would be particularly limiting if there is a bottleneck where they would just line up to be slaughtered by AoE attacks before they can deal damage.

Dugy:
I think that now there is an unintentional new problem - if two units with thorns attack each other, they will keep reflecting damage until one of them dies.

white-haired-uncle:
There can be only one.

white-haired-uncle:
I'm not seeing it.  If thorns used harm_unit_loti, sure.  But harm_unit doesn't trigger attacker_hits (like h_u_l)?

Dugy:
Okay, it didn't kill because it didn't award experience, and it didn't award experience because it needed `[harm_unit_loti]` to give experience on kill but not on every instance of dealing damage. So I put `[harm_unit_loti]` there and changed the way how it triggers events.

For completeness, I have also added abilities _reflect aura_ and _weak reflect aura,_ that reflect damage based on damage dealt if adjacent units are hit.

white-haired-uncle:
reflect_aura?  Man, that's just mean.  Basically the only way I ever lose units is if I don't notice a demon with reflect until it's too late.

Will I be able to tell if an enemy unit is affected?  Seems like something on the order of an illuminates/darkens "field" is warranted here.

Dugy:
I didn't add it to any enemies, not even as a demon trait, exactly because of this reason. However, 1.18 allows adding halos to abilities. I have accepted the pull request (#650 ) that added them to the usual leadership that had them added in quite a hacky way, adding them now should be significantly easier.

white-haired-uncle:
One more thing on whirlwind, since I'm thinking about it.  There are some units like the elves in the SW of Gods who transform into something else on "death".  But not when they're killed by ww, not even if it's a secondary.  Not a big deal in Gods, HUGE in (I think) Wasteland_of_Lost_Eternity.

Dugy:
Do you have a save file where it can be replicated? Getting there takes some time.

white-haired-uncle:
Target the leader.

[Uploading LotI2-Where the Hatred Grows Turn 2.gz…]()

Dugy:
I have added visual clues to the reflect abilities (will also apply to demons). The reflection aura should be very apparent.

Btw, you haven't uploaded the save file, only a weird link to this same page.

white-haired-uncle:
Sorry, link fixed.

If the reflect aura only has visual for demons, I suspect it's only to be given to demons?

Will there be more thorn/reflect-like auras with different damage types so they can be matched to unit types (fire for drakes, impact for trolls, etc)?

Dugy:
> If the reflect aura only has visual for demons, I suspect it's only to be given to demons?

No. I think you misunderstood my comment. Both reflects and reflect auras now have a visual, regardless of the bearer. This is mainly a QoL improvement that highlights reflect demons, but it is applied to every unit with the ability.

white-haired-uncle:
I just crafted Marrowrend for E (anger), which led to a thought.  What if lethargy for whirlwind (cleave, explosive, maybe blizzard, etc) was based on number of attacks attempted?  Or base plus attacks.  Seems logical (you work harder so you should tire more) and might have a desirable effect.

Dugy:
Oh, Marrowrend + whirlwind, I didn't realise that these two make an absolutely overpowered combo.

The choice to adjust the amount of lethargy stacks according to the number of attacks seems tailored exactly for this issue, and would lead to optimisations to trade the number of attacks for damage. In an extreme case, the Touch of Death gloves could be used to reduce the penalty very close to zero.

An additional problem I see here is that the penalty for AoE attacks reduces the chance that enemies suicide during their turn and make space for more enemies, making it paradoxically easier to continue blasting.

This leads me to an idea that it might be better to reduce resistances rather than damage.

white-haired-uncle:
> The choice to adjust the amount of lethargy stacks according to the number of attacks seems tailored exactly for this issue

Not just this one.  Sometimes you use a whirlwind-type attack on one unit, say because you don't want to get hit, or for the right damage type, etc.  And maybe it only takes one or two attempts (an attempt being one hit or one miss, not separate combats) to kill him.  That shouldn't be as tiring as using it against four enemies while making eight attempts at each.

Of course there needs to be a base cost as well,

Oh, and I just got temptationed by a brat in Into the Shadows.  Not trying to argue about it, but you mentioned it was rather specific but I do run into it in unexpected places.

Dugy:
And thanks for the nicely prepared save file for the enemies-not-reviving-when-killed-by-AoE problem. It turned out that the problem went much deeper, and it was a thing probably long before the recent change.

Dugy:
I have changed the resistances of Preserved Liches to be more akin to 1.18 lich. This makes them less vulnerable to arcane, but less resistant to blade and pierce. Efraim has worse resistances to be similar to elusive units (60% defence on flat), so in the end, he has only 20% base resistance to pierce. This should make him more vulnerable to thorns.

white-haired-uncle:
I don't think it will make any difference.  I think the problem is with absorb/suck.

When I advance E, I don't care inconsequential things like Meteor or Particle Storm, I concentrate on what really has an impact.  Things like +1 melee.  And in particular I always go for health/regeneration first, to get to absorb ASAP (with a little faster thrown in so he can fly).  E is going to give and take a ton of hits more turns.  Even something so seemingly trivial as absorb(1) makes a huge difference, much more than any "super weapon" attack.  Same for my gear choices, it's all about the number of hits.  

I also like anything that gives him more hits since they will be multiplied by furious.  More hits, more chances to heal.  Though I recently stopped giving him the blue(?) orb from the dragon (long scenario, awesome for taking on bzb, can't remember the name, Lost maybe?).  

Let's say I have the usual absorb(2) and a suck(1).  With thorns only doing 4 pierce, I need only 25% resistance to negate it completely.

I haven't looked at the details, but I suspect an approach would be to limit absorb to defense only, and suck to offensive hits on primary targets.  Making it once per attack instead of once per hit would be interesting.  Adding lethargy to each absorb/suck would be quite interesting, turning these from make-me-invincible to keep-me-from-dying specials.

On a somewhat related note, negative lethargy could (should?) affect movement as well.  I'm not sure how much, whether it would act as a percentage like slow or just -1MP for each N lethargy, but if you're tired you should also move slower. 

And I suppose wrath should cost as well.  A temporary boost in strength should result in a need for an eventual resting period.  Perhaps each +4 from wrath costs -1/4MP.  Then when you use wrath to offset lethargy (like I do almost constantly), you pay a double penalty -- ideally using whirlwind with a lot of hits against 3-4 enemies and wrath to offset the  lethargy should leave you pretty much unable to move.  I'm not sure on the details here, just making it up as I go, but I think there's a good idea in there somewhere.

Dugy:
The point of thorns is not to deal unhealable damage, but to add a contributing factor that could make it impossible to keep a positive total health gain. In case of thorns/reflect auras, it makes AoE more affected than other attacks. Thorns damage does not trigger absorb, by the way.

> Adding lethargy to each absorb/suck would be quite interesting, turning these from make-me-invincible to keep-me-from-dying specials.

This seems interesting. I will think of it.

> On a somewhat related note, negative lethargy could (should?) affect movement as well. I'm not sure how much, whether it would act as a percentage like slow or just -1MP for each N lethargy, but if you're tired you should also move slower.

I don't think this would be particularly effective, the enemies would just come to him.

> I haven't looked at the details, but I suspect an approach would be to limit absorb to defense only, and suck to offensive hits on primary targets.

Absorb is more useful on defence than on offence. Suck might be useful on both, yeah.

white-haired-uncle:
> The point of thorns is not to deal unhealable damage, but to add a contributing factor that could make it impossible to keep a positive total health gain. In case of thorns/reflect auras, it makes AoE more affected than other attacks. Thorns damage does not trigger absorb, by the way.
> 

Oh yeah, forgot.  It was leeches, not absorbs.  Either way, he heals completely.  I'm not really worried about him making positive health gains because he always starts with full HP.


> > On a somewhat related note, negative lethargy could (should?) affect movement as well. I'm not sure how much, whether it would act as a percentage like slow or just -1MP for each N lethargy, but if you're tired you should also move slower.
> 
> I don't think this would be particularly effective, the enemies would just come to him.

It makes logical sense (you're exhausted), and it negates penetrates/push so you can't use to smash your way forward.  Also, if implemented as -MP/-lethargy, you would remain less able to move for a turn or so (the MP penalty wears off with lethargy, not next turn like slow).

> 
> > I haven't looked at the details, but I suspect an approach would be to limit absorb to defense only, and suck to offensive hits on primary targets.
> 
> Absorb is more useful on defence than on offence. Suck might be useful on both, yeah.

I might mean leeches, not absorb.  The point is to take away some of the opportunities to heal caused by having so many hits (given and/or taken), but from AoE and furious/berserk.

white-haired-uncle:
E takes on a boss dragon with reflect.  And takes 3 damage.  This is not using the new resistances, not sure if they changed for blade, but it's a good example of where I think we'd like to see him at least get scratched.  It's not an AoE (I used furious), but I think it demonstrates the problem I'm having with him being invincible.

[LotI1-First Confrontation.gz](https://github.com/Dugy/Legend_of_the_Invincibles/files/14858189/LotI1-First.Confrontation.gz)

Dugy:
So I played that scenario to the end with your units. I lost two units because I was rather out of focus and exposed them unintentionally, but you had many good units who could cut through the army very well. Argan also almost got burned to ash because of his low ranged damage and low fire resistance. The other preserved liches were doing much better, enemies usually didn't dare to attack them, but they were taking some damage with the updated resistances.

Efraim's build, on the other hand, was too good. Even with updated resistances, his health was going down only when he was fighting the big dragon. The AoE attacks were not particularly powerful, but the lesser berserk was unstoppable. I don't know how you want to solo that, there isn't enough turns to kill all the drakes with leaders only. Did you really get through it leader-only by advancing one hex per turn? If yes, then the problem isn't what I thought it was and reflect won't help much there, because even if it might hurt a little, it isn't very consistent, and also you may avoid attacking most enemies with reflect.

Overall, it seems like absorb, suck, leech and murderlust can combine to provide enough healing to compensate for sustained low damage indefinitely. And you have stacked so many defences on him that he has both high resistance and low chance to get hit, so most enemies can't deal enough damage to outdo the heal.

white-haired-uncle:
It's not just that scenario.  Other than a handful of scenarios that are pretty nasty (mostly ch10 at this point), it's pretty much always like that.  I have to be careful with him for the first few scenarios, but after that...

That used to be one of the hardest scenarios, until I learned not to advance my army to the river, but leave everyone but E/L back at the castle.  You only have to kill the leaders, so just carve through the drakes using penetrates, skip the dragon and kill the other leader (the one that recruits), then mop up the remaining drakes for XP.

In this case, you're seeing a test version of the scenario where I've done a lot to try to stop that strategy.  I used a micro AI to force all Sky Drakes and Burners to attack canrecruit=no, made the corner into a poisonous swamp, and placed a bunch of drakes at the start to try to get in E/L's way so that the dragon has time to recruit (normally I'm within a few tiles of the enemy castle by turn 3 or 4.  Other than the fact that it's hard to place drakes in the right position where either E or L (E in this case) can't run around them early on, I'm pretty happy with the changes, they seem to put the army at risk, and while it's still not hard by any means you do have to try a little if you don't want to lose anyone.  I think it still needs a bit more work, but I'm thinking it might be time to release it into the wild and see if we get any reactions.

BTW, part of the reason it's so hard to block E/L is that I emphasize speed.  There are a lot of scenarios where I think I can see how you intended things to go, but with E and/or L at speed 10 or higher I can just run around danger, or rush the enemy castle before they have a chance to recruit.

> Overall, it seems like absorb, suck, leech and murderlust can combine to provide enough healing to compensate for sustained low damage indefinitely. 

Exactly.  I don't go for big damage per hit, I go for lots of hits, small amounts of healing PER HIT, and big resistances such that E/L pretty much never take damage.  I've always taken harder to hit early, which is good when he is weak, but since you're eventually fighting enemies that almost exclusively have magical I'm thinking that might be a waste.

As I've mentioned, I'd like to see some of those suck/absorb/etc tuned down so they don't apply to like every hit in combat.  I'm also thinking that absorb should come much later in the AMLA path.  Perhaps not let suck stack on hard.

I went ahead and played it out.  Finished on turn 12 with no losses.

Scenario and replay attached in case you want to look at anything.

[LotI1-First Confrontation replay 20240403-201428.gz](https://github.com/Dugy/Legend_of_the_Invincibles/files/14859964/LotI1-First.Confrontation.replay.20240403-201428.gz)
[global_events.cfg.gz](https://github.com/Dugy/Legend_of_the_Invincibles/files/14859991/global_events.cfg.gz)
[13_First_Confrontation.cfg.gz](https://github.com/Dugy/Legend_of_the_Invincibles/files/14859992/13_First_Confrontation.cfg.gz)

Dugy:
> It's not just that scenario. Other than a handful of scenarios that are pretty nasty (mostly ch10 at this point), it's pretty much always like that. I have to be careful with him for the first few scenarios, but after that...

In that case, it's serious.

> That used to be one of the hardest scenarios, until I learned not to advance my army to the river, but leave everyone but E/L back at the castle.

I managed that with your army using a fairly regular strategy. The 2 losses I had were due to lack of attention rather than a necessary sacrifice.

> In this case, you're seeing a test version of the scenario where I've done a lot to try to stop that strategy.

I am not sure about that, it's a start scenario save file and those don't snapshot the scenario code. Though I was a little surprised that leaders were not attacked, I explained that with their high damage and defence because Delly wasn't attacked either.

> BTW, part of the reason it's so hard to block E/L is that I emphasize speed. There are a lot of scenarios where I think I can see how you intended things to go, but with E and/or L at speed 10 or higher I can just run around danger, or rush the enemy castle before they have a chance to recruit.

Hm... I get that, but reducing speed after AoE attacks would not help in this case.

> As I've mentioned, I'd like to see some of those suck/absorb/etc tuned down so they don't apply to like every hit in combat. 

The problem I see with that is that their individual effects are minuscule. They are so efficient only if they all happen at the same time and are combined with good defences and resistances. There is no individually overpowered mechanic here, or any overpowered interaction.

This makes me think of a solution they took in Path of Exile, I game I often play and it has a lot of complex character mechanics. It used to be possible to stack massive amounts of armour to get such a massive damage resistance to weaker hits that it was possible to ignore weaker enemies completely (but armour was less effective against bigger hits, so it was often necessary to get a lot of it, and the devs couldn't just reduce the effectiveness). So they came up with a mechanic that each hit temporarily reduced armour by a tiny bit and lots of weaker could eventually reduce it to values where their damage started to matter.

I wonder if it wouldn't be of help to make armour lose effectiveness a little bit after every hit, so that getting swarmed would be a problem even for the best equipped units. It could be explainable as the armour getting worn up and requiring repairs.

white-haired-uncle:
Hmm, armor losing effectiveness.  Seems like it would require a lot of work on our end, since an item (on unit,ground,inventory) is no longer just item[57], it's now item[57][x], a specific instance of and item.  Not impossible, and leads to some interesting use cases, like limits (just how many swords did Eidolon have?).

But, without re-implementing all the behaviour of an object in "user space", how would this work with wesnoth's "static object" model?  I'd hate to have to have Efraim rebuild everytime his armour took a hit.  I guess you could make the damage rare and limited (0.01% chance per hit to take damage, each time armour is damaged it is weakend 25% -- using numbers I just made up).

I still really like this idea.  Even without item damage, it led me to some other thoughts like putting a cap on instances of an item.  For example, on normal there's only one Doombringer at any time, if you already have one another cannot be dropped unless you lose it.  On hard, there's only ONE Doombringer, lose it (or it gets worn out?) and it's GONE.

Dugy:
> I'm sorry, I posted the wrong save. I meant to post turn 9, where E is adjacent to the dragon, so you could see that he takes down the dragon (with reflect) in one turn while taking only 3 damage.

Can you just upload a save file to a place where you wanted to demonstrate that? Both global_events.cfg and the scenario content are snapshotted into mid-scenario save files, so you don't need to bother with supplying those. Usually, only lua is not snapshotted, and unit properties and items will persist until inventory is manipulated or the unit advances (except advancements, those are snapshotted forever).

> Agreed. That would help in some cases, like when I fly around casting blizzard to slow large armies (think Uria and her mechanical toys) every turn (assuming the speed reduction is not implemented using slow). 

Interesting idea. It will have to be taken into consideration.

> Defense/resist certainly offset a lot of the "reflect" damage. But it's the healing, IMO, that sets it to ZERO. And N*0=0 no matter how large N (hits) gets.

The point of reflect was to cause users of AoE attacks to take more damage and have an additional problem with sustaining their use. Against individual enemies, it's just a minor contribution that can offset healing effects, but if only few units have reflect it won't do much.

One thing that could make it more significant is to ignore the resistances when reflecting. To take part of the damage the enemy took. Otherwise it can be trivialised by maximising resistance to one's own attacks, as you did.

> Lose the word temporary, and I love this idea. It wouldn't even have to be "lose effectiveness", just give some items a life. For example, Iron Armour can block N damage, or take M hits. It will do so at full effectiveness, but when it's used up it breaks and cannot be repaired. Same for weapons. Add remaining life (as percentage) as a visible attribute to an item. Having the item lose effectiveness as it wears out would be even better. This would address a couple things I can think of

This seems too extreme to me. I was only hoping to prevent individuals from fighting for too long even if they can heal it back.

> Luck. Finding Searing Ice, or even Doombringer (which probably should be relegated to one of a kind status) on the first scenario, and the game is completely different. 

There is a pool of the strongest items that can drop only from very limited sources. Doombringer is one of them. For that reason, no enemy will drop it in the first scenario. It can be randomly found as a hidden treasure in _The Awakening,_ though. Usually, they can be found in hidden spots in dungeons. Overall, their main source is Beelzebub.

Overall, I think that it would be possible to make them even more scarce.

> I usually have several hundred unused items in my inventory by the end of ch5/9, and over 100 at the end of 4/8. It's nice to be able to play with all of the combinations, but items completely lose the rarity that makes them special (I think chapter 5 and to some extent 9 are supposed to be a way to collect a lot of loot, but I usually have so much going in there's little point in picking things up). 

The possibility to break them into gems was added to deal with the excess of boring items. At a later point, I have made it possible to adjust the distribution of items you find. So it's totally possible to make the more interesting items more scarce, making boring items common enough to use them on throwaway units.

> Hmm, armor losing effectiveness. Seems like it would require a lot of work on our end, since an item (on unit,ground,inventory) is no longer just item[57], it's now item[57][x], a specific instance of an item. 

Actually, I had thought that over. Its resistances could be applied as an invisible ability and their effect would change with a variable set inside the unit. This would allow taking damage to gear without recalculating stuff too often. The resulting durability of the item would be saved in the inventory, but permanent damage to gear feels like a last resort solution to me.

> I still really like this idea. Even without item damage, it led me to some other thoughts like putting a cap on instances of an item.

That would make some sense for super rare items, but you might just end up collecting them all by farming Beelzebub or something, and then the drops would be significantly affected by that.

Dugy:
Sorry, I forgot about this:

> Unrelated: I found out the very, very hard way recently that fast_ai is enabled by default for ALL scenarios. While it does seem quite important for performance, it might explain why the AI is so stupid most of the time, like sending 100 imps at Efraim in one turn (or why allies attack demons that regrow more then the ally could do in damage???).

That micro AI was implemented specifically because of LotI and I participated in its design (though it was written by the guy who wrote the other micro AIs). It was originally part of LotI, but then it was made part of the core game. The problem with it is that LotI has units with many attacks and a lot of abilities and weapon specials, which makes AI calculations extremely expensive. The default AI works as something like this pseudocode:

```C++
while (a_unit_can_do_something()) {
  float best_attack_score = -9000;
  attack_action best_attack = {};
  for (unit own_unit : this_side.units()) {
    for (unit enemy : enemy_side.units()) {
      attack_action attack(own_unit, enemy_unit);
      float attack_score = expensively_calculate_action_score(attack);
      if (attack_score > best_attack_score) {
        best_attack_score = attack_score;
        best_attack = attack;
      }
    }
  }
  perform_action(best_attack);
}
```

Micro AI fast does this pseudocode instead:

```C
for (unit own_unit : this_side.units()) {
  float best_attack_score = -9000;
  attack_action best_attack = {};
  for (unit enemy : enemy_side.units()) {
    attack_action attack(own_unit, enemy_unit);
    float attack_score = expensively_calculate_action_score(attack);
    if (attack_score > best_attack_score) {
      best_attack_score = attack_score;
      best_attack = attack;
     }
   }
  perform_action(best_attack);
}
```

It will do a lot more dumb things, because it will never consider if there is a better unit to attack some target, but it can handle massive armies with lots of units with lengthy individual damage calculations.

> P.S. I searched, but I can't find it. Do you do something in later chapter 5 to heal all units at the start?

No idea. I was writing that over 10 years ago and I didn't change it much since then.

white-haired-uncle:
> The point of reflect was to cause users of AoE attacks to take more damage and have an additional problem with sustaining their use. Against individual enemies, it's just a minor contribution that can offset healing effects, but if only few units have reflect it won't do much.

The only reason I ever lose any recalled/geared units is reflect.  I send a prophet who can do like 110x7 against a demon, not realizing that it has reflect, and, well it doesn't go well.  I find it to be the single most powerful tool the enemy has.

> One thing that could make it more significant is to ignore the resistances when reflecting. To take part of the damage the enemy took. Otherwise it can be trivialised by maximising resistance to one's own attacks, as you did.

I always thought it did ignore resistance until I looked at it recently.

> There is a pool of the strongest items that can drop only from very limited sources. Doombringer is one of them. For that reason, no enemy will drop it in the first scenario.

 I assume you're referring to Doombringer, item 801.  Doombringer item 72 doesn't seem to be very rare.

> Actually, I had thought that over. Its resistances could be applied as an invisible ability and their effect would change with a variable set inside the unit. This would allow taking damage to gear without recalculating stuff too often. The resulting durability of the item would be saved in the inventory, but permanent damage to gear feels like a last resort solution to me.

I'm not sure I follow all of that, but the very first thing I thought of when I started thinking about damaging the armor on a unit was, "what if he just drops it and picks it right back up?".

> Can you just upload a save file to a place where you wanted to demonstrate that? 

[LotI1-First Confrontation-Auto-Save9.gz](https://github.com/Dugy/Legend_of_the_Invincibles/files/14878600/LotI1-First.Confrontation-Auto-Save9.gz)

> It will do a lot more dumb things, because it will never consider if there is a better unit to attack some target, but it can handle massive armies with lots of units with lengthy individual damage calculations.

My concern is it's turned on for every scenario, unless you turn it off.  Which turns out to mean it's turned on for every scenario.

Dugy:
> The only reason I ever lose any recalled/geared units is reflect. I send a prophet who can do like 110x7 against a demon, not realizing that it has reflect, and, well it doesn't go well. I find it to be the single most powerful tool the enemy has.

Because of retaliation?

> I always thought it did ignore resistance until I looked at it recently.

Maybe it's broken. If the setting of damage is empty because of accessing an empty variable, it will ignore resistances.

> I assume you're referring to Doombringer, item 801. Doombringer item 72 doesn't seem to be very rare.

Doombringer 72 is not supposed to ever drop. It's there just to avoid breaking old save files. If it dropped in the last year or so, it's a serious bug.

> I'm not sure I follow all of that, but the very first thing I thought of when I started thinking about damaging the armor on a unit was, "what if he just drops it and picks it right back up?".

Items that aren't on units are saved with number and type (because of crafting). Durability could be stored with it as well.

However, as I said, I'd prefer to make rare items more scarce and instead make the armour damaging temporary, and turn it into a unit-specific feature rather than item-specific. This would keep the damage despite changing inventory.

> [LotI1-First Confrontation-Auto-Save9.gz](https://github.com/Dugy/Legend_of_the_Invincibles/files/14878600/LotI1-First.Confrontation-Auto-Save9.gz)

I see what you did there.

> My concern is it's turned on for every scenario, unless you turn it off. Which turns out to mean it's turned on for every scenario.

The slowdowns caused by the normal AI were a very serious annoyance, it was almost a no go.

white-haired-uncle:
> Because of retaliation?

Yes.  We've seen that it doesn't do much against a lot of little hits when you have some resistance and basically self-healing, but then you attack with a unit that does massive damage the retaliation can be fatal.

> Doombringer

All I know is it's not very rare.  Recently (I thought it was Where the Sun Does Not Shine, when the dragon was killed), I actually picked up two of them practically on top of each other.  Maybe that's a fluke extremely rare event, but it brings me back to my original point, a single instance of luck finding a great item at an opportune time can have a huge impact on the rest of the game, almost(?) to the point of ruining it (if you're not one that just likes to run around and have every enemy fall down on their own swords in fear when you approach).  Having items wear out and/or break seems like a fair way to offset this.

(Or maybe it is very rare, but since there's a HUGE number of drops it might need to be more rare.)

> I see what you did there.

That phrase is often used to say, for example, you caught me trying to sneak something past you.  There shouldn't be anything sneaky about that example.  You hit the dragon "boss", killing him easily, and take all of about 3 damage from his reflect after all of your resistance and healing.  The save should be "clean", though I played that scenario in bits and pieces a couple hundred times trying to get the AI working the way I wanted, the right drakes in the right place to slow down E/L's advance, etc, so it's possible there is cheating in that particular example.  Still it should make the point however.  (I should note he's still got the old resistances, and probably carries an old Godless, but I doubt those will make much difference).

Dugy:
> All I know is it's not very rare. Recently (I thought it was Where the Sun Does Not Shine, when the dragon was killed), I actually picked up two of them practically on top of each other. 

That looks like a bug that needs further attention. In Part 1, Doombringer was supposed to come from only 2 sources:
1. items found in hidden locations in caves (which often go unnoticed even after multiple replays)
2. Beelzebub (his drops need to be OP because he's there to discourage farming enemy recruits until the last turn)

Is Doombringer the only item like this? Are there some other items that seem totally OP in part 1 or too abundant for their strength?

> That phrase is often used to say, for example, you caught me trying to sneak something past you. There shouldn't be anything sneaky about that example. You hit the dragon "boss", killing him easily, and take all of about 3 damage from his reflect after all of your resistance and healing. 

I mean that you ran around the army, enganged them lightly with your units, while you cut through the mass on the recruit points to get to the dragon over several turns. This trick wasn't supposed to work, but it describes the exploit you were describing.

white-haired-uncle:
> > All I know is it's not very rare. Recently (I thought it was Where the Sun Does Not Shine, when the dragon was killed), I actually picked up two of them practically on top of each other.
> 
> That looks like a bug that needs further attention. In Part 1, Doombringer was supposed to come from only 2 sources:
> 
>     1. items found in hidden locations in caves (which often go unnoticed even after multiple replays)
> 
>     2. Beelzebub (his drops need to be OP because he's there to discourage farming enemy recruits until the last turn)
> 

On my current run I also picked up Doom in SW corner of Where the Sun Does Not Shine (just one this time).  I'm pretty sure last time the dragon dropped one (if that's possible) and there was one just next to where you kill him. 

I doubt I've ever went through both parts without finding Doom.  Probably not even one part.  I'd say considerably more than 50% of the time I get more than one per part.  I'm sure it helps that I pretty much know where all the loot is.  I do wish there was only one, and it could get broken -- I guess I can just smash it from time to time.

> 
> Is Doombringer the only item like this? Are there some other items that seem totally OP in part 1 or too abundant for their strength?

The Searing Ice is pretty gnarly.  Incinerate+slow, two different (probably) additional damage types, suck2 for the other, and a couple of +15 resistances.

Soul Hunter (murderlust) is kind of crazy for E, or it would be if he actually needed it (wouldn't be so bad if murderlust was offense only).

Warheart and Hatred if you stack them.  Probably Dying Dreams as well.

BTW, E currently has Mental Sheath and Draingloves, both absorb1, but they don't stack.  Suck stacks, and I'm pretty sure you can stack absorb (learned) with absorb (item special).  It's confusing.

Dugy:
> On my current run I also picked up Doom in SW corner of Where the Sun Does Not Shine (just one this time). I'm pretty sure last time the dragon dropped one (if that's possible) and there was one just next to where you kill him.

I looked at the scenario code and it looks like _Where the Sun Does Not Shine_ is a scenario where super items can be found in hidden corners. It's possible they are not so well hidden because the scenario was created before the items were added. And the dragon also drops 1 super item (I didn't notice this previously). Either way, finding it twice at almost the same spot is possible only in this scenario and also requires some luck (each time, the probability was slightly below 5%).

Doombringer 72 is not supposed to drop anywhere, unless you have a very old version.

> I doubt I've ever went through both parts without finding Doom.

That is quite likely because Beelzebub drops 19 items, all of which are from the superb item drop pool. Which means the probability of getting a Doombringer from a Beelzebub kill is 45%.

> The Searing Ice is pretty gnarly. Incinerate+slow, two different (probably) additional damage types, suck2 for the other, and a couple of +15 resistances.

Yeah, I guess this could be completely removed from part 1.

> Soul Hunter (murderlust) is kind of crazy for E, or it would be if he actually needed it (wouldn't be so bad if murderlust was offense only).

And it's a rather cheap craft. But there are also other relatively easy sources of murderlust, and I know the ability was always used frequently. Maybe it's time to replace the easiest sources of murderlust with some kind of _lesser murderlust_?

> Warheart and Hatred if you stack them. Probably Dying Dreams as well. Wretched Defiler. Really anything that adds +1 attack probably shouldn't add more than one resistance/damage/penetration/etc (and by one, I mean one resistance OR one damage bonus OR ..., not one of each).

Yes, that makes sense. Those dwarves with hammers dealing 35-5 come from these, right?

I wonder, should these be made weaker or moved into the super item drop pool? Or split into two, leaving a weaker one as a crafting recipe and moving the original into the super item drop pool?

> BTW, E currently has Mental Sheath and Draingloves, both absorb1, but they don't stack.

It is an event-based ability, and these don't stack with themselves. Only the last absorb applies.

> Suck stacks

It does, because it has code that sums up the values and applies them as a weapon special later.

> I'm pretty sure you can stack absorb (learned) with absorb (item special).

I would be surprised if it did.

white-haired-uncle:
> And it's a rather cheap craft. But there are also other relatively easy sources of murderlust, and I know the ability was always used frequently. Maybe it's time to replace the easiest sources of murderlust with some kind of lesser murderlust?

My only real concern with murderlust is E.  There are turns when 30-40 enemies attack him (and die).  The way I build him murderlust is probably unnecessary since he'll heal most/all of the damage anyway, but murderlust times lots of kills in one turn is a problem, IMO.  Less so for L, maybe that's just how I play but she doesn't kill nearly as many in retaliation.  For normal units that might kill a unit or two per turn I don't see a problem with it.

>>    Warheart and Hatred if you stack them. Probably Dying Dreams as well. Wretched Defiler. Really anything that adds +1 attack probably shouldn't add more than one resistance/damage/penetration/etc (and by one, I mean one resistance OR one damage bonus OR ..., not one of each).

> Yes, that makes sense. Those dwarves with hammers dealing 35-5 come from these, right?

Probably.  +1 attack is my go to, particular for high damage / low attack units.  And berserk/furious.  And trickery/doom/etc.

> I wonder, should these be made weaker or moved into the super item drop pool? Or split into two, leaving a weaker one as a crafting recipe and moving the original into the super item drop pool?

The problem with +1 attack, even if you cut out most of the other benefits from that item is that you can stack them (and combine with anger).  I think the way to do this is to give all +1 attack items a penalty, _and make it the same for all of them_.  For example, give all of them -2 or -3 MP, and if I try to stack 3 of them it's almost impossible for me to offset that with.  [I would also add a penalty to Mastadon, since crafting a few of those for one higher level unit gives crazy results.  And it's not just high level, I sometimes craft a set of Mastadons so I can give them all to a unit who I want to protect as I level it up.  -1 attack (melee and ranged) would be interesting]

I'm also a fan of sets.  I only know of one crafted set, a set of mixed crafted and dropped items might be interesting.  If you take Wretched Defiler and make everything but +1 attack and maybe plague dependent on a guantlets item then I can't combine it with Quick Gauntlets unless I want to give up the good damage and great arcane defense.  I prefer powerful items to have drawbacks, but breaking them into sets could work.

And of course borking +1 attacks items increases the relative value of the Blue Dragon Orb.

Dugy:
> Less so for L, maybe that's just how I play but she doesn't kill nearly as many in retaliation. For normal units that might kill a unit or two per turn I don't see a problem with it.

So a _lesser murderlust_ that doesn't work on defence?

> I'm also a fan of sets. I only know of one crafted set, a set of mixed crafted and dropped items might be interesting. If you take Wretched Defiler and make everything but +1 attack and maybe plague dependent on a guantlets item then I can't combine it with Quick Gauntlets unless I want to give up the good damage and great arcane defense. I prefer powerful items to have drawbacks, but breaking them into sets could work.

This is an interesting idea.

> And of course borking +1 attacks items increases the relative value of the Blue Dragon Orb.

That is kinda okay, since Blue Orb of Dragonflame is available only once, and is less useful for the leaders.

Dugy:
I have turned the +1 attacks on _Warheart_ and _Hatred_ to set effects requiring gloves that don't have +1 attacks. I have turned _Dying Dreams_ into a hard to obtain helmet and replaced the crafting recipe with a similar item without _suck_ and +1 attacks but with the other properties made stronger.

Let us continue the balance discussion in the issue you created for it.

PR #717, opened 2024-03-27
keep your wolf alive

white-haired-uncle:
Now that the wolf can recruit, let's move the castle tiles between him and Umbra's side.  Also make him a Great Wolf, and give him a tiny income.  This is only going to keep him alive for a couple turns, but at least he won't die immediately.  He must have been put there for a reason.  I suspect if I knew that reason I'd do more to help the little guy.


Dugy:
Thanks.

PR #719, opened 2024-03-27
invalid recruit= entries

white-haired-uncle:
Given the context and the scenario, I'm reasonably sure these just shouldn't exist at all.


Dugy:
All right.

PR #727, opened 2024-03-30
clear damagers variable

white-haired-uncle:


Dugy:
I wonder how much clutter was this adding to save files...

white-haired-uncle:
Probably right about none.  But if I happen to see something while I'm working on something else...

I had a 4KB external RAM expansion pack for my first computer.  I have to restart wesnoth regularly with 1.19/loti on my 4GB laptop lest I lock it up.  No idea where the OOM killer has gone.

Dugy:
No, I meant that these little sources of clutter accumulate and the save files are large enough to cause short but noticeable freezes when they're written. This one could have contributed some 1% to the overall size, which is not so small.

I have already added some code that specifically reduces the size of LotI save files, for example walking corpses with no items are purged from the recall list automatically.

white-haired-uncle:
Well, if you get really bored some time, I couldn't find what leaves updated[] and items[] around (tutorial is probably the simplest example).  I'm pretty sure items could just be cleared with a global scenario_end event.  Not sure about updated.

Dugy:
Those seem like some leftovers from times before a lot of item-related mechanics were moved into lua.

PR #728, opened 2024-03-30
 attribute to QUANTITY = HARD

white-haired-uncle:
When not clear, I assumed QUANTITY was correct.


Dugy:
Yes, those were set incorrectly.

PR #729, opened 2024-03-30
add kill= to harm_unit_loti

white-haired-uncle:
Why do I always look for loti_harm_unit?


Dugy:
> Why do I always look for loti_harm_unit?

If I make a LotI-specific edit of something, I always use the campaign shortcut as suffix, never as prefix.

PR #731, opened 2024-03-31
Fix syntax for textdomain binding

reportingissueinprogress:


reportingissueinprogress:
This scenario is currently not being detected by wmlxgettext due to a wrong syntax in the textdomain binding.

PR #732, opened 2024-03-31
improve recall old units

white-haired-uncle:
When you recall units from the past (Invasion, Bloodbath, etc) you only get a list of name/type.  It's hard to remember which is which unless you named them just right.  This adds a unit_preview_pane so you can see a lot more info about the units you are considering.

I gave it it's own recall.lua.  I have more recall gui related features in progress, so this isn't (won't be) just a one-off.

tested on 1.19-dev


Dugy:
Oh, I didn't know there was a `unit_preview_pane` widget. Cool.

PR #733, opened 2024-03-31
Item list

white-haired-uncle:
Also includes cleanup in tutorial

I hacked test_scenario to skip item 57 for now.

Set to draft as I've noticed Invasion is taking about 60 seconds to go from story to scenario when started from scenario start.  Need to revert and see if this is related.


white-haired-uncle:
Ok, it's not the recent changes.  Invasion just sucks.  Here's the problem:

```
        {VARIABLE recall_count 200}
        [while]
            [variable]
                name=recall_count
                greater_than=1
            [/variable]
            [do]
                {VARIABLE_OP recall_count add -1}
                [recall]
                    [filter_wml]
                        [variables]
                            mass_produced=yes
                        [/variables]
                    [/filter_wml]
                    x,y=36,7
                [/recall]
            [/do]
        [/while]
        {CLEAR_VARIABLE recall_count}

```
Every recall takes about 1 second.  I tried make a stored list of all mass_produced and then recall by id, same thing.

Ideas?  Maybe store them at the end of earlier scenario instead of putting them on the recall list?

Move this to start so the player can at least see something happening?

Dugy:
I think that the problem is that `[filter_wml]` is very slow, with a few exceptions. One of them is when it's checking something inside of the unit's `[variables]`. I suppose this exception only applies while the game is set up.

It's fine if you move it to scenario start.

white-haired-uncle:
I thought it would be filter_xml, and I wondered if it was expanding every unit on recall each time it recalled one unit.  So I moved the filter to store_unit so it would just happen once, but the recalls were still slow even without the filter_xml.  Oh well, works now.

Discontinuum:
I'm a bit out of context, why do we want to delete GENERATE_ITEM_LIST?

Dugy:
> I'm a bit out of context, why do we want to delete GENERATE_ITEM_LIST?

An independently generated item list exists in lua and almost all item-related code is in lua now. Duplicate information makes bugs worse and is a waste of resources.

PR #735, opened 2024-03-31
performance fix and tweaks

white-haired-uncle:
Lines 105-108 in scenarios4/11_Invasion.cfg don't belong.  I pushed a second commit to clean them up, but I have no idea where it went.  EDIT: never mind, fixed


Dugy:
Thanks.

PR #742, opened 2024-04-01
order of events for DROPS

### LINKED ISSUE #741: DROPS_EXT
In The_Real_Enemy there are lots of drops, even though the chance is low and you don't kill that many units.  While investigating, I ran into this.  Is this redundant, or was something else intended?  I'm a bit dense today, but this doesn't look right to me.

```
        [if]
            [variable]
                name=unit.canrecruit
                equals=yes
            [/variable]
            [and]
                [variable]
                    name=unit.canrecruit
                    equals={BOSSES}
                [/variable]
            [/and]
            [then]
                {DEFAULT_DROP_ITEM $x1 $y1 {WEAPONS}}
            [/then]
        [/if]

```

white-haired-uncle:
Closes #741

### DIFF
--- a/scenarios5/06_The_Real_Enemy.cfg
+++ b/scenarios5/06_The_Real_Enemy.cfg
@@ -216,7 +216,7 @@
         [/endlevel]
     [/event]
     {BEELZEBUB_SPAWN_POINT 3 1 18 8 13-28,3-13}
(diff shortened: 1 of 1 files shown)


Dugy:
Oh, cool. I didn't expect this, but there were so little things specific to this scenario that there weren't many other places where this could be.

PR #748, opened 2024-04-07
have no_fast_ai accept a side list

white-haired-uncle:
I want to disable fast_ai, but I only need to do so for side 3.  Also tested "2,3".  "yes" is still all sides.


Dugy:
Considering the agonising delays caused by the default AI's handling of larger groups of complex units facing each other, it's better to use it only under very specific circumstances, like if the enemy has only very few units.

PR #752, opened 2024-04-08
add hills/cave to lesser nightstalk

white-haired-uncle:
Rockbound Cave (Hills, Cave) - should be included since it's hills?


Dugy:
Yeah, should have been included.

PR #756, opened 2024-04-10
Make E work harder for super furious attack

white-haired-uncle:
I never take "more precise with sword".  Why would I, furious comes with marksman?  Wait, what?  Furious comes with marksman???

This is an untested markup of how I'd like to see marksman applied to furious.  Or something in the neighborhood.

Also, some comments on naming (berserk 1-2-3 = mberserk2-3-5), a question on how damage works if you take it before say berserk2 (which should be berserk3?).

I'm not sure that damage shouldn't have the same progression (3 AMLAs to get extra damage on mberserk3, not 1).

Looks like it's not only untested, it hasn't been through a spell check.


Dugy:
I am fine with the changes. You can even remove the marksman for the 5 repeats version if you want.

white-haired-uncle:
I wonder what swarm would look like on furious.  Might reduce his chances to heal fully every turn (but perhaps only when he is seriously hurt), might not.  If it weakened him to the point where he didn't kill every single unit that attacked him, the ones that survived would block others from attacking him on the same turn (and being attacked is generally a good thing for E).

Dugy:
I think swarm would weaken him overall, because it would reduce his healing from hits and kills, allowing him to heal only from getting hit and attacking himself. Getting mobbed could make him take less damage overall, but with less possibilities to heal, so he would probably be able to survive it for a couple of turns, giving you time for rescue, but not indefinitely.

white-haired-uncle:
> I think swarm would weaken him overall, because it would reduce his healing from hits and kills, allowing him to heal only from getting hit and attacking himself. Getting mobbed could make him take less damage overall, but with less possibilities to heal, so he would probably be able to survive it for a couple of turns, giving you time for rescue, but not indefinitely.

I deleted my comment when I noticed that he has swarm on furious.  I guess I just never let him get hurt badly enough to notice.

white-haired-uncle:
I've been looking at this, and everytime sword gets marksman it applies to furious as well.  At first I thought they were tied somehow with force_original_attack=sword (undocumented), but that doesn't seem it be it. Also, apply_to=bonus_attack is undocumented.

I guess I can add an effect that removes marksman from furious when furious is created and when sword get marksman (since order is unknown)?  But perhaps there's a better way?

Dugy:
> At first I thought they were tied somehow with force_original_attack=sword (undocumented), but that doesn't seem it be it. Also, apply_to=bonus_attack is undocumented.

It's a LotI-specific feature. It clones an attack and makes the desired changes to it. It picks the best attack (damage * attacks) with the given range, unless `force_original_attack` is specified.

One way to make it more expensive would be to require going through a 4-berserk instead of going straight to a 5-berserk.

white-haired-uncle:
Are there things in bonus_attack that are needed for furious, or could it just be a new attack?

One option is just to remove marksman from furious, and then move (on higher levels) up in a the amla chain.  "more precise with sword" will still immediately affect all three furious levels, but at least it would come a lot later.  something like
`
{QUANTITY require_amla "sword1" "sword3" "sword5"}`

white-haired-uncle:
Or perhaps remove "more precise with sword" (higher levels at least) and only create "more precise with sword when attacking twice/etc" as separate entities.  I don't think I ever use sword once I have furious.

Dugy:
> Are there things in bonus_attack that are needed for furious, or could it just be a new attack?

Yes, it could be a new attack, and it that case, it could have separate upgrades than the normal sword attack. But I would not be against removing it completely.

white-haired-uncle:
I've made the furious attacks into new_attack, created separate AMLAs for marksman, and placed a bit more dependency on taking 'better with sword' multiple times before you can advance them all.  Also, marksman for fury(5) is only available on EASY.

I've tested the dependencies, but of course the overall impact is yet to be seen.  For me, the biggest impact will probably come from new_attack, as I don't think furious attacks will inherit slow from Doombringer anymore.  Fury(2) can do more damage than the standard sword attack, but with the right specials I suspect there may be times when players would actually choose the sword instead of just picking fury(5) every time.

I also moved plague so it is much easier to get so players can use it to build undead earlier in the game (I never take it, as it comes too late to be of any use).

Dugy:
The problem with this approach is that no weapons apply to that berserk attack, which would look like a bug to just about everybody.

white-haired-uncle:
You mean the weapon-specific specials that won't be "inherited" by being a clone of the sword, or that it will say Fury (2) - blade instead of Konrad's Might - blade or Flaming Sword - fire ?

white-haired-uncle:
Screenshot_2024-05-08_05-49-31

Dugy:
> You mean the weapon-specific specials that won't be "inherited" by being a clone of the sword, or that it will say Fury (2) - blade instead of Konrad's Might - blade or Flaming Sword - fire ?

The latter. It doesn't count as a sword attack in that case, which means that weapons you are carrying won't apply to it.

white-haired-uncle:
That's the way it's worked as long as I can remember.

The snapshot is from the way it worked without the changes.

Dugy:
> That's the way it's worked as long as I can remember.

I understand it's good for balance, but I am quite sure people would keep reporting it as a bug.

white-haired-uncle:
I don't understand.  Do they report it now?

Dugy:
No, but I am sure they will start once this is released.

white-haired-uncle:
Why? It's not changing.  See the screenshot from BEFORE the change. It's "always" said "fury (2)", not "Sword of Krux".  

The difference they'll see now is they won't get leeches for furious attacks and they won't get marksman until they advance it.

Dugy:
Yes, because bonus attacks usually override attack name and icon. But if it doesn't inherit weapon specials from the main attack, it will look like a bug.

I would suggest to add some undesirable weapon special to those fury attacks, such as -10% chance to hit.

white-haired-uncle:
I don't like -10% chance to hit, as it conflicts with marksman/magical etc.  Instead, I added a 10% chance to take some self-inflicted damage (when you're attacking furiously you tend to be a little less careful than normal).

Fury is a bonus_attack again, but each furious level takes more AMLAs to attain than before, and precise takes a lot of "better with sword" (since it will affect all fury attacks as well).  Plague comes earlier.  Marksman removed until earned

If we come to agreement on this I'll push a clean version when done.

Dugy:
10% chance to injure oneself is an option, but it seems a little unexpected, there is no self-injury yet and I wonder it it wouldn't be exploitable by triggering something.

How about a +20% chance to get hit by the opponent?

white-haired-uncle:
Chance to be hit has the same problem, there's too many enemies with magical.  And of course E doesn't care about getting hit since he heals from pretty much everything (that's not self inflicted).

I used to take 'lower chance to be hit' heavily very early in my AMLAs.  Recently I've started ignoring them since they become worthless.  I may rethink that, since late in the game all AMLAs are pretty useless anyway, and lower chance to be hit is rather useful for a few scenarios when E is pretty fragile.

Dugy:
So how about making him unhealable during the fight?

white-haired-uncle:
That could work too, assuming absorb/suck/leech/drain/whatever all honor unhealable.

I'm kind of partial to the idea of hasty, where because you're attacking really fast you tend to be kind of careless and can injure yourself.  

I also like the idea (for other attacks) of clumsy, where you (sometimes) injure adjacent allies with your AoE attacks.  Clumsy would also be good for some really OP attacks, like a prophet with charge and backstab and a bunch of damage -- make it a little dangerous to use, at least around your friends.

And of course, I'm pretty convinced at this point that whirlwind (and probably cleave) should ALWAYS injure allies.  Magic AoE attacks that only affect enemies I can buy, but spinning around like a madman with a sharp weapon and NOT hitting EVERYONE next to you?  Nah.

Dugy:
I am not much in favour of self-damage and friendly fire, mainly because I remember this being a thing in the _Path of Exile_ game and it was always allowing all kinds of crazy builds based on positive feedback loops.

> That could work too, assuming absorb/suck/leech/drain/whatever all honor unhealable.

It should work because all heals are based on `[heal_unit]` that is from core. In case it doesn't work, it can be fixed.

Dugy (approved):

white-haired-uncle:
It looks like leeches still leeches, absorb doesn't.  Need to check drains and sucks.  It's hard to keep them separate, particularly since he had leeches inherently from the looks of it.

white-haired-uncle:
I'm thinking 'furious' would be a better name than 'hasty'.

But then when I think about whether this should apply to WCT and maybe any lesser berserk, I wonder if a separate special is needed at all -- non-healing could just be rolled into lesser berserk.

white-haired-uncle:
Doing more damage with furious - looks like it adds 10%, though it doesn't look like it by the syntax?  Really should be +1 IMO, adding 2-3 damage on an attack with a potential minimum of 5x5 attacks is pretty extreme.

white-haired-uncle:
Made second_unit undrainable during the fury attack, and filtered leech so it doesn't activate with special hasty.

Unless cleave is an issue, I think this should do it if we only want this to affect Efraim.

Dugy:
> Made second_unit undrainable during the fury attack, and filtered leech

I am still surprised that leech ignores unhealable and the others abilities respect it, all of them are based on `[heal_unit]`.

Drain would still ignore it without making the opponent undrainable.

white-haired-uncle:
Yeah, I just looked at it and I don't see why leeches would be any different.  But it kept doing it even when he was unhealable.

Dugy:
So absorb was suppressed and leeches wasn't?

white-haired-uncle:
The way I remember it, absorbs and sucks were blocked by unhealable, while drains and leeches were not.  Then I made the enemy undrainable, which stopped most of the "healing" except for a rare +1 (which I assumed was the cumulative effect of leeching fractions) which was eliminated by filtering out the hasty special for the leeches events.

BTW, while looking at this, I ran into this.  I think the 'not_living' needs to be 'undrainable'.  Shouldn't be related, but worth looking at.

```
                       # Note: Damage that is too small to leech can be combined from multiple targets into a real hitpoint.
                        # Example: 4 damage to 3 targets. Each hit alone produces 0 leech gain, but all 3 together get 1 hitpoint.
                        [if]
                            [variable]
                                name=has_leech
                                not_equals=0
                            [/variable]
                            [then]
                                {VARIABLE_OP damage_to_leech add $true_damage}
                            [/then]
                        [/if]
                        # Drain
                        [if]
                            [variable]
                                name=this_item.status.not_living
                                not_equals=yes
                            [/variable]
                            [and]
                                [variable]
                                    name=has_drain
                                    not_equals=0
                                [/variable]
                            [/and]
                            [then]
                                {VARIABLE_OP damage_to_drain add $true_damage}
                            [/then]
                        [/if]

```

Dugy:
I looked at the code of `[heal_unit]` and it does not respect `unhealable` in any way. Now I am surprised that absorb is not suppressed.

>  I think the 'not_living' needs to be 'undrainable'.

Yes, in the past, there was a status named `not_living` and it meant simultaneously undrainable, unplagueable and unpoisonable. This was quite inconvenient, mainly for various eras featuring vampires that were meant to be immune to drain and plague, but not poison.

white-haired-uncle:
absorb has animate=no.  It may be still working, I just don't see it happen.

Dugy:
> absorb has animate=no. It may be still working, I just don't see it happen.

Very likely.

white-haired-uncle:
Suck/absorb/leech now filtered for attack.special=furious, so no unhealable needed.

undrainable now uses [set_status_semaphore].

pre_advance_stuff or advance_stuff should be updated to clear variables used by [set_status_semaphore] now that it exists.  It's not necessary for this case, but in case it's ever used for something else.

Efraim_lich is working.  Once you're okay with his changes I will apply to Efraim_god.

Vritra and Vritra_later also have fury, but I'd be very concerned about weakening her in ch10.

Question about whether to apply to other lesser berserk attacks is still open.

Hmm, if I advance E's fury sufficiently, and then give him WCT....

Dugy:
I didn't think much about exploits based on Wrathful Combat Techniques. My typical workflow was to add quite random things and see if it got overused. I haven't seen overuse of Wrathful Combat Techniques.

I can change `[heal_unit]` to respect unhealable to help you finish this.

white-haired-uncle:
> I didn't think much about exploits based on Wrathful Combat Techniques. My typical workflow was to add quite random things and see if it got overused. I haven't seen overuse of Wrathful Combat Techniques.

I know I've posted a few recently.  I believe one was Manta+blue orb+WCT+anger=100 attacks.  Oh, 100 attacks WITH WRATH.
5x attacks + wrath after only 4 AMLAs is pretty exploitable.  Might make sense to move lesser berserk to a different book (and have it grow more slowly), and add different stuff to WCT (triggerable-like stuff?).

Actually, I should quit going after damage and build a unit with a ton of attacks and trickery, doom, mayhem, lethargy, etc.  I always try to keep a unit with many hits and trickery around for BZB, but I've never gotten crazy with it.

> 
> I can change `[heal_unit]` to respect unhealable to help you finish this.

I'm thinking about requesting honor_unhealable=yes|no (default no) to be added to [heal_unit] (could be my first wesnoth PR, should be pretty straight-forward).  I think a [heal_unit_loti] could also come in handy so you had control over other things (a remove incinerate argument comes to mind).  But for this project, I've got it working by adding a filter to absorb, leech, and suck, so [heal_unit] doesn't come into play.  The only "complicated" part was parry, and that would need doing no matter how [heal_unit*] works.

Dugy:
> I know I've posted a few recently. I believe one was Manta+blue orb+WCT+anger=100 attacks. Oh, 100 attacks WITH WRATH.

True, but remember only you experimenting with that.

> 5x attacks + wrath after only 4 AMLAs is pretty exploitable. Might make sense to move lesser berserk to a different book (and have it grow more slowly), and add different stuff to WCT (triggerable-like stuff?).

How about simply putting there some rather weak advancements and then just berserk 2, or berserk 3?

> I'm thinking about requesting honor_unhealable=yes|no (default no) to be added to [heal_unit] (could be my first wesnoth PR, should be pretty straight-forward).

Yes, but that would not be backported to 1.18. To make it work in 1.18, I would have to use some monkey patching.

white-haired-uncle:
> How about simply putting there some rather weak advancements and then just berserk 2, or berserk 3?

That would be better than it is now, though any combination of wrath and berserk is probably too much for one item/legacy/etc IMO.  

Actually, (lesser) berserk might be limited to a potion.  That just seems like something temporary that you would get from a potion, not permanent like you would learn from a book.  Maybe stackable books, so each lesser B potion added 1x to your attack(s).

Or perhaps WCT gives wrath and berserk2 after a couple weak amlas, WCT Vol II gives berserk3 (only, and one with WCT Vol I).  Or something.

I better stop, I just started thinking about craftable potions.

white-haired-uncle:
```
[heal_unit]
   honor_unhealable=yes|no

```
I suppose that's just 
```
[heal_unit]
   <whatever>
   [filter_wml]
       [not]
          status=unhealable
       
```

white-haired-uncle:
Just saving this for reference.  My first try in part2 using the harder to get fury attacks (on normal diff), I finally got fury5 in Gates of Hell.  Looks like I may have wasted a few AMLAs chasing the dark dragon (they were useful at the time), but other than that and the necessary flying and absorbing I concentrated on sword attacks to get to this point.

E_part2_fury5_path

Dugy:
Do you still need the change to make `[heal_unit]` respect the unhealable status?

white-haired-uncle:
> Do you still need the change to make `[heal_unit]` respect the unhealable status?

No, though I still think there will be a [heal_unit_loti] some day.

Dugy (approved):

Dugy:
I have made `[heal_unit]` respect unhealable in LotI code, unless it has an attribute `respect_unhealable=false`.

PR #757, opened 2024-04-10
Ch3 make hard harder

white-haired-uncle:
Progress:  My first BZB kill finally came in Pits, taking advantage of the terrain.  My army is weaker than usual (good), though some of my units are starting to become quite powerful by the end of the chapter.  Given that it could be luck, or my knowing how to play the changes I'm making, I'm going to assume things are going about as expected and not try to adjust the scenarios any more than if I were doing a little less well (as I had hoped).

I'm not sure I quite understand this chapter.  It's got the autorecall, no upkeep (or maybe it's just all the loyals), no early finish, 100% carryover style I've seen before, but there's also a few early finish and I assume a few where your gold is added to a scenario starting gold.  Even after lowering the hidden gold, I still find that I end up with more gold by the time I get to the trader in Bitter Swamps than I started with.  I like the idea of having a trader near the end to give the player a reason to be frugal, but 2,000+ gold to spend there is too much.  And, as always, no gold carryover between chapters on hard.

FWIW, I usually spend my gold:

1.  S1 a couple fighters, a couple archers
2.  a couple scouts at some point
3.  one or two heavies
4.  maybe a couple of dwarves, but just because I have a lot of axes and I'm not short on maces
5.  a few mages
6.  no bandits
7.  some recalls for BZB, a few for the dwarf optional mission


white-haired-uncle:
Most of the chapter is the usual removing L2+ from recruit list, some gold/income/turns adjustments, and terrain changes generally to protect enemy leaders from quick kills.

Notable:

Lost:  The Dragon.  Normally, I take out the drakes and then head his way, E&L alone while the rest of the army collects XP.  I find the dragon and trigger the dialog.  One turn to approach.  One, MAYBE two turns to kill him and take his orb.  The blue orb is way too valuable (and getting better with +1 attack item adjustments) to not have to fight for it.

I want to make it much more of a fight, and to require more than just E&L (E or L really).  My plan to do this is to surround him with powerful support such that E/L can't just fly in and smash him in a turn. To do so:

1. Make the lava slow side 1.  Basically, take away E/L's penetrates.
2. Make the dragon and adjacent units very resistant to arcane and blade, E/L's bread and butter.
3. AI tweaks to ensure his army actually protect him.
4. Oh, and a LOT MORE HP.  To get the blue orb should be a boss fight, like Abbadon in Inferno at least.  [speaking of whom, I had a fun idea I might use here, similar to Verm Sewers.  Give the dragon rotating auras.  Like the way Abby can only be hit by one damage type per turn, but in this case mostly effecting his army.  Later, maybe]

I also replaced some of the rare items to ensure the player has tools so that they can defeat the dragon.  Not easily I hope, but it shouldn't be the case that the orb is unretrievable just because they didn't happen to develop/find any cold attacks yet or something.

First Confrontation:  We saw this earlier, just try to make it so E/L can't charge ahead through the drakes.  [I suspect the fast_ai is part of the problem, I read somewhere that it puts undue focus on attacking leaders which may be why the drakes just stand around in a huge pack instead of finding someone to kill.

Before I figured out you can do this, I used to find this scenario challenging.  Obviously, you head for the river and fight like heck with everything you have.  I'd like to bring kind of that fight back.

Chasing Dragons:  This one has always been a big let-down.  100 turns, yet I tend to finish in 12 (my best is 7).  With your flanks and rear covered and allies that can soak up most of the initial enemy wave (while E/L carve up dragons), I don't see much to do here.  I messed with resistances, since that's about the only thing I can think of that gives the enemy any sort of chance.

I'm reasonably happy with ch3.  The big three scenarios probably need more work.  I think they are much better now, but need to be tested by people who don't already know the right way to play them.

scenarios3/08_Lost.cfg
```diff
@@ -321,14 +378,35 @@
...
+        {PLACE_ITEM 18 87 55}  # Ring of Iced Veins - cold damage, slows and fire resist, just the thing if you want to take on a fire dragon
+        {PLACE_ITEM 97 87 68}  # Boots of Fleeing - I could use those to approach the dragon's castle!
+#ifndef HARD
+        {PLACE_ITEM 259 82 66} # Stolen Potion - skirmisher?  I could use that to approach the dragon's castle!
+        {PLACE_ITEM 527 77 69} # Hey!  Look over here!  Over here, next to the opening to the area with the hidden loot.
+#endif
         {RARE_ITEM 20 53}
-        {RARE_ITEM 87 55}
         {RARE_ITEM 47 47}
-        {RARE_ITEM 87 55}
```
Dugy, line 331: The goal was to replace 2 of the super rare items with items that are helpful to complete the scenario but not particularly unusual?

scenarios3/10_Gehenna.cfg
```diff
@@ -233,8 +233,12 @@
...
     [event]
         name=turn 2
+#ifdef EASY
         {RARE_ITEM 21 3}
+#endif
+#ifndef HARD
         {RARE_ITEM 22 3}
+#endif
         {RARE_ITEM 23 3}
         {RARE_ITEM 21 2}
```
Dugy, line 243: Can you please replace them by less valuable items instead of removing them entirely? It sucks to discover a hidden room and find nothing in there.

scenarios3/10_Gehenna.cfg
```diff
@@ -275,16 +279,22 @@
...
         [/filter]
+#ifndef HARD
         [sound]
             name=gold.ogg
         [/sound]
+#endif
         [message]
             speaker=unit
+#ifdef HARD
+            message= _"It's empty!"
```
Dugy, line 290: It would be hilarious to add a remark that someone had embezzled their entire treasury and that it's a shame that you can't just wait for them to go bankrupt.

scenarios3/14_Chasing_Dragons.cfg
```diff
@@ -325,19 +326,20 @@
...
+#endif
     [/side]
     [side]
         type=Orcish Warlord
         id=warlord
         generate_name=yes
         random_traits=yes
         side=4
-        canrecruit=yes
+        canrecrut=yes
```
Dugy, line 337: Typo?

scenarios3/14_Chasing_Dragons.cfg
```diff
@@ -749,6 +852,8 @@
                                     message= _ "Something happened."
                                 [/message]
                                 {EARTHQUAKE (    {MODIFY_UNIT side=2 status.petrified no})}
+                                # on hard, consider adding objective here to keep side 2 leader alive.  He shows no interest in self-preservation, so it could be a challenge.
+                                # of course, on hard if you trigger these guys you should be ashamed of yourself.
```
Dugy, line 856: Might be worth the little change it takes.

utils/global_events.cfg
```diff
@@ -1992,3 +1992,111 @@ $soul_eating"
...
+            [filter_condition]
+                [have_unit]
+                    side=1
+                    [filter_location]
+                        terrain=Ss,Ss^*,Chs,Khs,*^Vhs
+                    [/filter_location]
+                    [filter_wml]
+                        [status]
+                            poisoned=yes  # Note: you won't get second warning until a unit that is poisonable spends a second turn in swamp.
+                        [/status]
```
Dugy, line 2033: Can this cause performance slowdowns on new turn in practice? `[filter_wml]` is claimed to be quite slow unless used on the unit's variables.

Dugy (changes requested):
A few minor remarks.

scenarios3/14_Chasing_Dragons.cfg
white-haired-uncle: yes, I fixed a bunch of those the other day, but I must have been working from an old copy.  Or I just missed one completely.

utils/global_events.cfg
white-haired-uncle: I don't know.  I would think that the first two filters would have to match (side 1 unit in swamp) before the third would be evaluated, and then hopefully only for units that matched the first conditions.  And I read that filter_wml does not expand units unless necessary (is it for status?).  I just copied most of this from AE's Dead Menance.  I certainly didn't notice any slowdown in the scenarios where it was used.

scenarios3/10_Gehenna.cfg
white-haired-uncle: I can, but it's not that there's nothing in there.  There's still 4 items on easy, 3 on normal, and 2 on hard.  And a chest.

Dugy: Aha, yeah, sorry, I didn't notice they were all close to each other.

utils/global_events.cfg
Dugy: If you didn't observe anything, it's probably fine.

Dugy:
Good ideas with the dragon.

Regarding Chasing Dragons, I was thinking that if it is a large scenario where the leaders can avoid the enemy army to go for the leaders, maybe the leader could react to that situation (2 leaders nearby, few units around them) by summoning two very fast and powerful monsters who would each on one leader, with abilities focused on countering its target.

> I read somewhere that it puts undue focus on attacking leaders which may be why the drakes just stand around in a huge pack instead of finding someone to kill.

It looks like it focuses even on leaders it cannot reach, screening other units, right?

white-haired-uncle:
> > I read somewhere that it puts undue focus on attacking leaders which may be why the drakes just stand around in a huge pack instead of finding someone to kill.
> 
> It looks like it focuses even on leaders it cannot reach, screening other units, right?

It seems that way to me. Like it doesn't understand, or doesn't care, that an enemy leader which is one turns movement away but blocked by AI units it "farther away" than an enemy unit it can move to in two turns.  Of course, that may not be just fast_ai.

I do feel like the AI in loti plays pretty stupid, placing way too much emphasis on attacking E/L (even when it's laughably suicidal).  I suspect it's the fast_ai, but I can't prove it (yet).

I sure know the difference when I turn it off.  All my hardware is extremely low end, but I doubt that's it.

BTW, I threw together a rudimentary loti-preferences menu (I hate the blood rain animation, especially on large maps when it gets intense).  An option to allow users to toggle the use of fast_ai seems like a decent idea, some have fast processors or maybe would rather play a slower but smarter AI (on some maps at least?).  Of course, in one of the ch3 scenarios I rearranged the order of stages specifically to keep fast_ai while also making other AI changes but that's based on the scenario controlling the use of fast_ai, not the user.

scenarios3/10_Gehenna.cfg
white-haired-uncle: try that

scenarios3/14_Chasing_Dragons.cfg
white-haired-uncle: ok, made it an option

white-haired-uncle:
While testing the changes to S14, I ran into something.  If you use the Ancient Necro at 33,34 to awaken the ancient warriors, and then refuse the deal 4 or 5 times she will be killed, turn into a lich, and disappear.  I'm not sure if this is the expected behaviour, I expected her to die and become a lich like she does in normal combat.

[LotI1-Chasing Dragons.gz](https://github.com/Dugy/Legend_of_the_Invincibles/files/14942333/LotI1-Chasing.Dragons.gz)

I advanced her to Ancient Necro thinking I'd take a bunch of AMLAs, then let her die and become a lich and her upgrades would carry over.  Bummer.  And of course she's stuck with legacy of light.

scenarios3/08_Lost.cfg
white-haired-uncle: Yes.  They're not great items, but they aren't bad. I'd rarely let slow and penetrates sit unused on the inventory list.  The RoIV also sets melee to cold, which negates suck, which keeps me from giving RoIV to E, which is probably a good thing.

white-haired-uncle:
BTW, Thirst for Adventures.  I can't speak to your state of mind at the time, but it looks like you wanted to assign the villages within radius 15 of side 2's keep to side 2.

```
        [store_starting_location]               #No idea what the hell is this - but it may have some purpose. Was I drunk when I was writing this?
            side=2
            variable=temp_starting_location
        [/store_starting_location]

        {CAPTURE_VILLAGES 2 $temp_starting_location.x $temp_starting_location.y 15}
        {CLEAR_VARIABLE temp_starting_location}
 
```

white-haired-uncle:
> Regarding Chasing Dragons, I was thinking that if it is a large scenario where the leaders can avoid the enemy army to go for the leaders, maybe the leader could react to that situation (2 leaders nearby, few units around them) by summoning two very fast and powerful monsters who would each on one leader, with abilities focused on countering its target.

I've hacked up something like this and now I'm working on customizing the units.  Do you know of a tag, or a macro something like

GIVE_UNIT_ITEM UNIT_FILTER  ITEM_NUMBER

I suppose it would also take SORT in case an item is craftable.

Probably make sense if there were others like GIVE_UNIT_ABILITY, TRAIT, etc, instead of doing that manually every time.

scenarios3/14_Chasing_Dragons.cfg
```diff
@@ -339,8 +339,8 @@
         recruit=Orcish Grunt,Orcish Archer,Orcish Assassin,Wolf Rider,Troll Whelp,Troll Rocklobber,Goblin Knight,Orcish Warrior,Orcish Crossbowman,Goblin Spearman,Goblin Impaler,Orcish Slayer
         team_name=good
         user_team_name=_"Good"
-        {GOLD 600 650 300}
-        {INCOME 20 30 10}
+        {GOLD 600 650 300}  # why is this higher on normal than easy?
```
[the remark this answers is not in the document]
Dugy: So it was probably following an incorrect assumption that it will matter. You don't need to keep it that way.

white-haired-uncle:
Okay, I think I'm close to what you wanted on Chasing Dragons.  I do need some text for a spell the dragon casts, and a name for the new attack (the idea is it shoots a lightning bolt from it's claws).

I also would like to see if [have_unit] could take radius= (x,y=starting_location(side)) instead of me passing in 23-45, 10-15 or whatever.

And it needs tested.  I added the knockback event because E/L get pretty busted up and I want the player to have a fair chance to try again and not just be killed because I don't like the way they chose to attack, but still those dragons are pretty nasty.

Dugy:
> Okay, I think I'm close to what you wanted on Chasing Dragons. I do need some text for a spell the dragon casts, and a name for the new attack (the idea is it shoots a lightning bolt from it's claws).

* Spell: _Thar'ta rargbta gorr'thon eibraj khabaragalarraaaa!_
* Attack name: Thunderclaw

> I also would like to see if [have_unit] could take radius= (x,y=starting_location(side)) instead of me passing in 23-45, 10-15 or whatever.

You have to put a `[filter_location]` into the `[have_unit]`, give it a radius and put a `[filter]` for the dragon into it.

white-haired-uncle:
>     * Spell: _Thar'ta rargbta gorr'thon eibraj khabaragalarraaaa!_

I'll make it translatable just to screw with people.

> > I also would like to see if [have_unit] could take radius= (x,y=starting_location(side)) instead of me passing in 23-45, 10-15 or whatever.
> 
> You have to put a `[filter_location]` into the `[have_unit]`, give it a radius and put a `[filter]` for the dragon into it.

I don't feel like screwing around with nested filter types that I barely understand.  Something like this?

```
    [event]  # Approached a dragon's keep with only E and/or L and saw what you get, now knock our heroes back out of danger (for now)
        name=side {SIDE} turn end
        id=heroes tried to solo side {SIDE}
        [filter_condition]
            [have_unit]
                side=1
                canrecruit=yes
                [filter_location]
                    radius=7
                    [filter]
                        id={DRAGON}.id
                    [/filter]
                [/filter_location]
            [/have_unit]

```

I seem to remember there's something important about where radius= goes (like before or after the filter perhaps).

Dugy:
Yes, I think that is correct.

> I seem to remember there's something important about where radius= goes (like before or after the filter perhaps).

No, the rule is that it's only accepted inside `[filter_location]`. The order of values inside a tag never matters (unless they have the same name, then one overrides the other).

white-haired-uncle:
The [have_unit] works, but this doesn't:

```
        [store_unit]
            [filter]
                side=1
                [filter_location]
                    radius=7
                    [filter]
                        id={DRAGON}.id
                    [/filter]
                [/filter_location]
            [/filter]
            variable=goodies
        [/store_unit]
 
```

Dugy:
How about `id={DRAGON}`?

white-haired-uncle:
I fixed that, and now it's the other way around.  store_unit works, have_unit doesn't

Dugy:
You've made it to work using `[store_location]`, it looks like. Is it ready for merge now?

white-haired-uncle:
Hmm, I wrote up a description of what happened, I wonder what happened to it.

The key was to make the knockback (turn end) event be created by the summon dragon event (turn start) so that an approaching dragon couldn't trigger only the later.  Using starting location was a troubleshooting step, but I kept it because it makes some sense and it's not going to make a difference most of the time anyway since they stick to their keeps.

I should probably test that one some more.  I'm tired of it, but it needs testing to make sure they're not too hard.

white-haired-uncle:
I'd change that last commit if I could.  Name should be "edit recruits" not "no gold carryover between chapters on hard".  Other than that, I'm good.

Dugy:
I changed it for you, squashed it and merged it.

PR #760, opened 2024-04-13
no cheating in tunnels

white-haired-uncle:
Chapter 5 tunnels were pretty cool, right about until I figured out cheating.  Leave a unit by the scenario entrance and as soon as you get into trouble (or have killed an enemy leader or two), cast soulgate twice, run back to the previous scenario (fully healed), cast soulgate twice, go back to the scenario you "failed", and try again (fully healed, with more XP, potentially a new autorecall list, and in some cases with one or more enemy leaders now defeated).

Obviously, soulgate needs a cooldown (changing scenarios NOT counted as a turn) and healing between scenarios removed (should be easy, just store your units, but they keep healing on me anyway).  But to being with I'm looking at an idea which is going to be a good bit of work so I'd like your opinion before I go too far with it.  Every scenario (save Entrance) has an objective.  You don't leave the room until you complete it, or they carry you out feet first.  Or perhaps, you can't leave the way you came in.  I'm starting with the later, as it provides the player an out should situations arise where it's hopeless (I assume there will need to be difficulty tweaks if you can't just Sir Robin at will).  

While I'd like to see this applied across the board, right now I only want to touch hard, and not just because the lower levels would require a ton of tuning.  Mostly, if a player runs into an impossible situation, they can always restart the scenario on normal and play on instead of having to wait for a fix.

While the file here works, it's not quite complete.  Mostly because if you invoke TRANSITION twice in the same place you have to answer "do you want to go..." twice.  Easy to fix, but I don't want to touch the macro until I know if I'm going to apply this to all 47 scenarios.  I included just enough to show intent.  [EDIT:  Actually, just give the event an id (which I'm a fan of anyway, since it makes debugging with inspect SO MUCH easier) and don't worry about it]

### DIFF
--- a/scenarios5/27_Power_Station.cfg
+++ b/scenarios5/27_Power_Station.cfg
@@ -57,6 +57,32 @@
         [recall]
             id=Lethalia
         [/recall]
+#ifdef HARD
+        [if]
+            [variable]
+                name=sc27
+                equals=1
+            [/variable]
+            [then]
+                # add check if already exist (put it in TRANSITION)
+                {TRANSITION 8-15,0-2 12,1 _"Entrance" 08_Entrance}
+                {VARIABLE transition_set[8] yes}  # move to TRANSITION
+                {TRANSITION 12-17,28-31 15,30 _"Amplificator 4" 22_Amplificator_4}
+                {VARIABLE transition_set[22] yes}
+            [/then]
+        [/if]
+        [switch]
+            variable=last_scenario
+            [case]
+                value=8
+                {TRANSITION 12-17,28-31 15,30 _"Amplificator 4" 22_Amplificator_4}
+            [/case]
+            [case]
+                value=22
+                {TRANSITION 8-15,0-2 12,1 _"Entrance" 08_Entrance}
+            [/case]
+        [/switch]
+#endif
     [/event]
     {RECALL_ALL}
     [side]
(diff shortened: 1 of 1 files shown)


white-haired-uncle:
This is the event that is restoring the hitpoints (which the wiki says modify_unit doesn't do).

`#define DISABLE_UPKEEP SIDE
    [event]
        name=unit placed
        first_time_only=no
        [filter]
            side={SIDE}
        [/filter]
        [filter_condition]
            [variable]
                name=unit.upkeep
                not_equals=loyal
            [/variable]
        [/filter_condition]
        [modify_unit]
            [filter]
                find_in=unit
            [/filter]
            [variables]
                original_upkeep=$this_unit.upkeep
            [/variables]
            upkeep=loyal
        [/modify_unit]
        [deny_undo]
        [/deny_undo]
    [/event]
`

Dugy:
Are you sure it's the `[modify_unit]` that does it? I am quite sure it doesn't.

white-haired-uncle:
Yes.  If I comment out {DISABLE_UPKEEP} the issue goes away.

And modifying the contents of modify_unit as above makes it act differently.

white-haired-uncle:
I went from Eastern Storeroom to Library to Eastern Guardroom, only to find you can't go to the Library from E.Guardroom.

Is this intended?  Or should there be no transition from Library to EG?

Dugy:
Library is a hidden location, so there aren't transitions to every nearby room.

white-haired-uncle:
This has been superseded.

PR #761, opened 2024-04-14
strength power generators

white-haired-uncle:
I was under the impression that the power generators were somewhat important, like they did something to weaken Akula.  It actually doesn't look like they do anything (at least not the ones in Power Station), but I already went into the work of beefing them up.  Might have to do something about that.  

Changed damage type to lightning, because it's called lightning and it makes sense.  Increased the damage, 6x3 isn't going to do much this late in part 1, even if it's lightning. 

Added full damage reflect ability.  Originally I wanted high damage when hit by metal weapons, but except for assuming all blade weapons are metal that's kind of hard to do.  So I just made it reflect on melee.  And it's lightning.  Seriously, don't hit stuff with sparks coming out.  This needs a halo.

I wanted to make it slow, so that units with penetrate couldn't back off to avoid retaliation (including improved deathaura), but that would also limit the amount of retaliatory damage.

Also, changed name of deathaura to deathaura (X) to make it easier to identify.

### DIFF
--- a/units/Power_Generator.cfg
+++ b/units/Power_Generator.cfg
@@ -6,7 +6,8 @@
     race=monster
     image="units/enemies/core-compact.png"
     ellipse="misc/ellipse"
-    hitpoints=80
+    hitpoints=720
+    {QUANTITY hitpoints 80 240 720}
     movement_type=smallfoot
     {TRAIT_MECHANICAL}
     [movement_costs]
@@ -21,7 +22,7 @@
         cold=70
         arcane=20
     [/resistance]
-    movement=7
+    movement=0
     experience=150
     level=3
     alignment=neutral
@@ -33,18 +34,34 @@
     description= _ "High amounts of an unknown kind energy are created in this machine, and used in an unknown way somewhere else. You have pretty much no idea what this thing is meant to do. The light it creates seems to harm everyone nearby."
     [abilities]
         {ABILITY_ILLUMINATES}
+#ifdef EASY
         {ABILITY_DEATHAURA 16}
+#else
+#ifdef NORMAL
+        {ABILITY_DEATHAURA 24}
+
+#else
+        {ABILITY_DEATHAURA 48}
+#endif
+#endif
+        {ABILITY_ELECTRICAL_DISCHARGE}
     [/abilities]
     [attack]
         name=lightning
         description=_"lightning"
         icon=attacks/lightning.png
-        type=fire
+        type=lightning
         range=ranged
-        damage=6
+        damage=18
+        {QUANTITY damage 6 12 18}
         number=3
         [specials]
+#ifndef HARD
             {WEAPON_SPECIAL_MAGICAL}
+#else
+            {WEAPON_SPECIAL_GUIDED}
+#            {WEAPON_SPECIAL_SLOW}   # want this to cancel penetrates, but don't want to affect attackers damage since we're going to return it
+#endif
         [/specials]
     [/attack]
     [attack_anim]
--- a/utils/abilities.cfg
+++ b/utils/abilities.cfg
@@ -1685,6 +1685,15 @@ Enemy units cannot see this unit during dawn or dusk, except if they have units 
         halo_image_self="halo/reflect-aura-1.png:100,halo/reflect-aura-2.png:100,halo/reflect-aura-3.png:100,halo/reflect-aura-2.png:100"
     [/dummy]
 #enddef
+#define ABILITY_ELECTRICAL_DISCHARGE
+    [dummy]
+        id=electrical_discharge
+        name= _ "electrical discharge"
+        female_name= _ "female^electrical discharge"
+        description= _ "When this unit is hit with a melee attack, the resulting discharge of electricity causes an equal amount of lightning damage to be taken by the attacker. It can kill."
+#        halo_image_self="halo/reflect.png"  # need something that looks like it's throwing off sparks
+    [/dummy]
+#enddef
 #define ABILITY_SOUL_EATER
     [dummy]
         id=soul eater
--- a/utils/abilities_events.cfg
+++ b/utils/abilities_events.cfg
@@ -6102,7 +6102,57 @@
             kill=no
         [/harm_unit]
     [/event]
-
+    [event]
+        name=attacker hits
+        id=electrical_discharge
+        first_time_only=no
+        [filter_second]
+            ability=electrical_discharge
+        [/filter_second]
+        [if]
+            [variable]
+                name=weapon.range
+                equal=melee
+            [/variable]
+            [then]
+                [harm_unit]
+                    [filter]
+                        x,y=$x1,$y1
+                    [/filter]
+                    amount=$damage_inflicted
+                    damage_type=lightning
+                    fire_event=yes
+                    experience=no
+                    animate=no
+                    kill=yes
+                [/harm_unit]
+            [/then]
+        [/if]
+    [/event]
+    [event]
+        name=defender hits
+        id=electrical_discharge
+        first_time_only=no
+        [if]
+            [variable]
+                name=second_weapon.range
+                equal=melee
+            [/variable]
+            [then]
+                [harm_unit]
+                    [filter]
+                        x,y=$x2,$y2
+                    [/filter]
+                    amount=$damage_inflicted
+                    damage_type=lightning
+                    fire_event=yes
+                    experience=no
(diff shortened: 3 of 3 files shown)


Dugy:
Akula's damage reflection and some other bonuses depend on the number of power generators left alive.

Either way, I like the idea that they can't be trivially killed, but the area has to be cleared before they are disposed of with safety.

white-haired-uncle:
It looks like power generators only effect spawning, and only slightly (clearing out the power generators is worth twice as much as collecting the items from a storeroom).

Would you like to shift their focus more toward the final battle?  Perhaps something like having the power generators supply northfrost aura to the enemy leaders?

The way it works now it would seem that as you clear more of the tunnels, they become easier to clear.  I'd like to see benefits from fulfilling objectives limited to the final battle for the most part.

Dugy:
The idea was that by causing enough damage to their infrastructure, their ability to field many soldiers will go down, allowing you to delve into harder areas.

Destroying Amplificators is supposed to primarily reduce Akula's reflection significantly, from prohibitive to annoying. Or 50 damage on hit isn't prohibitive?

white-haired-uncle:
> The idea was that by causing enough damage to their infrastructure, their ability to field many soldiers will go down, allowing you to delve into harder areas.

I didn't notice that there were harder areas.  I guess it's supposed to be progressively difficult as you get farther from the entrance (or close to the center)?.  Perhaps destroying the PGs helps you get to the more remote amps (2&3) which which then assist in weakening Akula (though 2&3 aren't really "remote" from amp1, you just have to go through a couple storerooms which are currently about as easy as it gets in ch5).

An alternative could be that PGs only affect the guardrooms (the 'inner circle'), giving an incentive to explore all of the outer circle first.  I prefer the progressively harder/deeper idea, but given the nature of the map it seems kind of hard to imagine.

Right now, the fight just gets easier and easier as you go (all rooms start off about equal, but as you clear them the spawning gets lighter, your units advance, and you gather tons of gear).

> 
> Destroying Amplificators is supposed to primarily reduce Akula's reflection significantly, from prohibitive to annoying. Or 50 damage on hit isn't prohibitive?

50 pierce/fire/etc (so like 20?), but only from some units and some attack types (I've only glanced at the details at this point) is mildly annoying, but I've never found it to be any hindrance to soloing her in less than 10 turns.  I have to choose my attacks a little more carefully, and maybe suck down some of those potions I never use in the rest of the campaign, but that's about it.  I'm not sure at this point, does whirlwind/AoE cause her/them to retaliate?  I seem to remember simply targetting adjacent enemies to dish out damage to her before putting her down for good.

Amplificators are next on my list.  IMO, if destroying an amp is going to knock down Akula's HP by 40% (that's gotta be adjusted for difficulty), or her attacks by 50% (again, adjust), it had better be at least as hard to defeat as say Abaddon in Inferno (not VS).  Right now, they drop easily in a single turn.  And most of them have castles, but don't recruit?

Dugy:
> Right now, the fight just gets easier and easier as you go (all rooms start off about equal, but as you clear them the spawning gets lighter, your units advance, and you gather tons of gear).

How about making the deeper rooms significantly harder, so that the infrastructure damage would be more important for progression?

>  I'm not sure at this point, does whirlwind/AoE cause her/them to retaliate? I seem to remember simply targetting adjacent enemies to dish out damage to her before putting her down for good.

It's damage reflection, so preventing retaliation doesn't help and now AoE attacks will not help avoid it either.

white-haired-uncle:
Okay, I see the comment in the SE storeroom about how the spawning is worst here because it's so deep.

But it's just spawning L3 and L5.  And some of those L5 (Scoundrel) are just L3.1 that grant more XP when you kill them.

I've added an option accelerator so that the spawn frequency increases progressively at some dynamically defined rate.  Initially the idea was you might set it when the room's objectives were cleared so players can't just farm XP/items (at least not so easily).  I think I'll play with that here.

white-haired-uncle:
I'm working on making each of the Amplificators and their teams have different strengths and weaknesses (amp3 has lava, so those units will be good with fire, bad with cold, affinity to lava, etc).   I'm playing with the idea of having a room where each side3 unit carries a craftable weapon.  Picking a random craftable weapon is easy enough, but I'm struggling with getting an enemy to take it.

loti.item.on_unit.add* seems obvious, but it's throwing me errors.  Not sure why yet.

Picking a random number and feeding that to a switch would work, but it's boring and crude.

Dynamically creating an [object] tag should be doable, and I though I saw a comment that this is already implemented, but I'm not sure how to put in in place.  By that I mean:

```
[modifications]
  {SOME_OBJECT_DEFINED_BY_A_MACRO}
[/modifications]

```
is easy, while:

```
[modifications]
  $variable_holding_object_definition_in_a_string
[/modifications]
```

is not.  I've been told [insert_tag] might be the way to go.

I know that many of the specials won't work for enemies, and they won't know how to intelligently apply what does work, but it's a fun little idea and I want to see how it works out.  I know Manta carries Dragon Claw, and you can decline her offer to join you and while I've never done so that makes me think this is at least possible.  Hoping you have an idea.

Dugy:
> Why is this the deepest room, because it's the farthest from Entrance, or the lowest (southernmost)?

Farthest from entrance, for gameplay reasons.

> But it's just spawning L3 and L5. And some of those L5 (Scoundrel) are just L3.1 that grant more XP when you kill them.

Feel free to make Scoundrel stronger. It could be an oversight.

> I think I'll play with that here.

Seems like a good idea.

> Picking a random craftable weapon is easy enough, but I'm struggling with getting an enemy to take it.

Picking up items is not meant for enemy units. It should work, in theory, but the effects might be removed when advancing and it hasn't been tested for bugs.

> loti.item.on_unit.add* seems obvious, but it's throwing me errors. Not sure why yet.

I think that changing the code not to throw those errors is the best approach here.

> I've been told [insert_tag] might be the way to go.

Yes,  `[insert_tag]` is what would do what you want here. It's what was used before `loti.item.on_unit.add` superseded it. However, the items as a WML array exist only to keep old code running (or maybe they were completely removed already, I am not sure).

> I know that many of the specials won't work for enemies, and they won't know how to intelligently apply what does work, but it's a fun little idea and I want to see how it works out.

This is not a significant issue, most of them would be annoying even if they weren't used efficiently.

What is a greater problem is that not all item properties are applied by the `[object]` being in `[modifications]`. Many of them need to apply only to specific attacks or in some specific ways, so they are implemented with special tags and properties that are read by LotI code, which handles their behaviour. For this reason, if you want all the properties to apply, you need to use `loti.item.on_unit.add`.

white-haired-uncle:
I ended up using [insert_tag].  The tag data (the item) is created by lua (has to be) and passed to wml in a variable.  And then [update_stats] is needed to make the effects apply to the holder.  I'm not overly worried about all the effects applying to an AI unit, it's basically a gimmick, a way to give one of the amplificators' minions some interesting random properties.

BTW, while trying to get loti.item.on_unit.add working I ran into a couple instances of wesnoth.deepcopy.  I couldn't find any documentation on this and didn't spend any time trying to figure it out, but a couple hours later I saw that it is broken in 1.18:

https://forums.wesnoth.org/viewtopic.php?p=688915#p688915

Dugy:
> I ended up using [insert_tag]. The tag data (the item) is created by lua (has to be) and passed to wml in a variable. And then [update_stats] is needed to make the effects apply to the holder. 

This should work, yeah. Maybe it's better this way, you avoid mixing WML with lua more than necessary.

I am surprised why `[update_stats]` worked on enemies but `loti.item.on_unit.add` didn't. Not sure if you were using it wrong or what.

Dugy:
Giving bosses a way to acquire berserk is not the smartest idea. It allows you to find a unit that the boss can't counter well, support him heavily, and take him down in a single attack.

Overall, I have a lot of doubt the AI can use all the abilities and weapon specials efficiently. Those that are not using events are considered by the AI, but it won't use them very strategically. Especially the fast ai that is actually capable of finishing the calculations on time.

Dugy:
>  I don't think the combat calculations that are supposed to simulate combat even do simulate combat (fully). 

They do simulate combat, but they don't consider the events that may be triggered by some weapon specials. Even the fast AI will simulate combat against every reachable unit for the unit it's moving and pick the one to whom it can do the most damage.

> I suppose I could make up a pool of weapons that only have AI-friendly features, but then I'd have to come up with names and such.

That is a reasonable way of doing it.

PR #773, opened 2024-04-21
EXPLOSIVE_SLOW description

white-haired-uncle:
Only works on offense.  Update description, and active_on so the gui will show it inactive on defense.


Dugy:
Good observation.

PR #777, opened 2024-04-23
Ch4 hard harder

white-haired-uncle:
Not a lot here.   Only interesting parts are Way of Assassins and maybe The Kingslayer.  They are now more about completing the mission than just cutting down everyone.

### DIFF
--- a/scenarios4/02_Into_Our_Frozen_Land.cfg
+++ b/scenarios4/02_Into_Our_Frozen_Land.cfg
@@ -7,7 +7,7 @@
     {SCENARIO_MUSIC "breaking_the_chains.ogg"}
     {EXTRA_SCENARIO_MUSIC "vengeful.ogg"}
     {EXTRA_SCENARIO_MUSIC "the_city_falls.ogg"}
-    {TURNS 30 27 25}
+    {TURNS 30 27 22}
     victory_when_enemies_defeated=yes
 
     {DEFAULT_SCHEDULE}
@@ -47,7 +47,11 @@
         canrecruit=yes
         side=1
         controller=human
+#ifdef HARD
+        recruit=Poacher,Thug,Footpad,Thief
+#else
         recruit=Poacher,Thug,Bandit,Trapper,Footpad,Outlaw,Thief,Rogue
+#endif
         {GOLD 500 450 400}
         income=2
         team_name=good
@@ -71,7 +75,8 @@
         canrecruit=yes
         recruit=Vampire Guard,Vampire Master,Vampire Master Bowman
         side=2
-        {GOLD 75 100 125}
+        {GOLD 75 100 150}
+        {INCOME 0 0 12}
         team_name=evil
         user_team_name=_"Evil"
     [/side]
@@ -82,7 +87,8 @@
         canrecruit=yes
         recruit=Vampire Duelist,Vampire Longbowman,Vampire Lieutenant,Vampire Swordsman
         side=3
-        {GOLD 125 150 175}
+        {GOLD 125 150 225}
+        {INCOME 0 0 12}
         team_name=evil
         user_team_name=_"Evil"
     [/side]
@@ -93,7 +99,8 @@
         canrecruit=yes
         recruit=Vampire Duelist,Vampire Longbowman,Vampire Lieutenant,Vampire Swordsman
         side=4
-        {GOLD 125 150 175}
+        {GOLD 125 150 300}
+        {INCOME 0 0 12}
         team_name=evil
         user_team_name=_"Evil"
     [/side]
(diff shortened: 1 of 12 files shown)


scenarios4/02_Into_Our_Frozen_Land.cfg
```diff
@@ -71,7 +75,8 @@
         canrecruit=yes
         recruit=Vampire Guard,Vampire Master,Vampire Master Bowman
         side=2
-        {GOLD 75 100 125}
+        {GOLD 75 100 150}
+        {INCOME 0 0 12}
```
Dugy, line 79: Add some income on normal as well, please. Other places as well.

scenarios4/13_The_Way_of_Assassins.cfg
```diff
@@ -192,9 +192,53 @@ A good plan. Let us go."
...
+                description= _ "Kill Lord Kartyn"
+                condition=win
+                [show_if]
+                    [have_unit]
+                        id=lord3
+                    [/have_unit]
+                [/show_if]
+            [/objective]
+            [objective]
+                description= _ "And then get Efraim to the castle on the eastern border of the map"
```
Dugy, line 222: The _and then_ connection is not needed here.

scenarios4/15_The_Library.cfg
```diff
@@ -606,7 +614,9 @@
             result=defeat
         [/endlevel]
     [/event]
+#ifndef HARD
     {DROPS 5 8 (sword,bow,staff,sword,bow) no 2,3}
```
Dugy, line 618: No DROPS defined on hard? I am not sure how well will it work.

scenarios4/17_Any_Means_Necessary.cfg
```diff
@@ -85,8 +85,12 @@
         canrecruit=yes
         side=1
         controller=human
+#ifdef HARD
+        recruit=Elvish Archer,Elvish Fighter,Elvish Scout,Elvish Shaman,Duelist,Horseman,Spearman,Fencer,Heavy Infantryman,Elvish Lord
+#else
         recruit=Elvish Archer,Elvish Fighter,Elvish Scout,Elvish Shaman,Elvish Marksman,Elvish Ranger,Elvish Hero,Elvish Captain, Elvish Rider,Elvish Druid,Elvish Sorceress,Lancer,Longbowman,Duelist,Javelineer,Pikeman,Swordsman,Dragoon,Shock Trooper,Horseman,Knight,Cavalryman,Spearman,Fencer,Heavy Infantryman,Elvish Lord
-        gold=1500
+#endif
+        {GOLD 1500 1500 700}
```
Dugy, line 93: 1200 gold on normal, please.

Dugy (changes requested):

scenarios4/13_The_Way_of_Assassins.cfg
white-haired-uncle: It's more accurate that way.  The other objectives can be completed in any order, but this one must come last.

scenarios4/02_Into_Our_Frozen_Land.cfg
white-haired-uncle: Ok, but I don't see the point.  In most cases, the additional income is needed to offset the cost of the AI recruiting/supporting higher level units.  In others, to avoid the classic 'fight off one big wave on favorable terrain and then advance and mop up'.  Since I'm generally not changing NORMAL and it was fine before with income=0 it should be fine no matter what changes on HARD.

scenarios4/15_The_Library.cfg
white-haired-uncle: You're already picking up like 6 books in less than 10 turns.

scenarios4/17_Any_Means_Necessary.cfg
white-haired-uncle: Done.  But again, changing normal only because hard changed?

Dugy: Yes. There should be some kind of progression, if a scenario is too easy on hard, it's probably too easy also on normal.

white-haired-uncle: Maybe so.  Or there wasn't much difference between normal and hard to begin with.

scenarios4/02_Into_Our_Frozen_Land.cfg
Dugy: But I think that Normal could benefit from some of these changes to some extent.

scenarios4/13_The_Way_of_Assassins.cfg
Dugy: Aha, okay.

PR #778, opened 2024-04-23
upgrade scoundrel

white-haired-uncle:
The Scoundel is a bit weak for a L5.  Upped his melee attack and HP a little, roughly matching the progression other units get for a two-level advancement.

Also, note the unit that's called Blackguard who is not a Blackguard.


Dugy:
Okay. I will renamed Blackguard to Knave.

Issue #787, opened 2024-05-08
Housekeeping - uneccessary code


white-haired-uncle:
I've been wondering just how much clutter is being carried around, stored in every save.  While I doubt it would make a noticable difference to anyone in terms of performance, memory, or disk space, I'm still curious.  This is a thread to consider related issues.  We got rid of item_list (in WML), what else is there?

The help menu is rarely used.  And it looks like its entire contents are stored in every save.  Should be easy enough to reconfigure that so that it is only loaded when accessed.

I'm looking at some saves from part1 which include a decent amount of redeem code.  If nothing else, this could be broken out so that it is only loaded in part2, and ideally only loaded once you achieve redeem (at which point, the soul eater code becomes stale -- I think).

I doubt any cleanup would provide much benefit, and trying to get to granular would probably just result in added complexity, but it's probably worth looking at just the low hanging fruit.

Dugy:
Smaller save files are more convenient to work with overall because text editors are a little slow from editing save files that are several megabytes large.

And yes, there is a lot of unnecessary stuff. There used to be much more of it, but now, a lot of stuff was moved to lua, so if I do random sampling of a save file's contents, it seems to be mostly campaign statistics, units mostly on the recall list, and scenario replay data. The units themselves take most space by far.

So if you want to move the help menu into a helper unit type or move the redeem code into lua, you can, but I don't think it's worth the effort.

The best way to reduce the save file size right now is to shorten the code of units. Removing all the properties of items except for number and slot is a low hanging fruit.

white-haired-uncle:
These are things from global_events which look like they could at least be only loaded in one part, not both:

        {REDEEMING}
        {LOTI_DEMON_HACK}
        {LOTI_MAKE_BLACK_SOUL}
        {LOTI_MAKE_DOPPELGANGER}
        {LOTI_KILL_ALL_DOPPELGANGERS}

Dugy:
I think its effect would be neglectful. Also, all of those except for redeeming are events used as functions, which should have been lua-based WML tags instead.

white-haired-uncle:
```
#ifdef CAMPAIGN_LEGEND_OF_THE_INVINCIBLES_PART_I
    {REDEEMING}
#endif
```

saves all of about 3K (30K uncompressed), which is probably less than 1% of most part1 saves.  Almost nothing for the rest.

[...]

PR #788, opened 2024-05-08
load help menu dynamically

white-haired-uncle:
Removes only about 32KB, but that's almost 9% of my early saves in Toxic Sun, or about 4% of those late in ch5.

It does make for a simple example of how to remove other stuff.

I did not remove the now unneeded utils/help.cfg.

#787

### DIFF
--- a/help/faq.cfg
+++ b/help/faq.cfg
@@ -1,0 +1,157 @@
+#textdomain wesnoth-loti
+[event]
+    name=help_faq
+    [message]
+        speaker=narrator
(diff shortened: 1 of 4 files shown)


Dugy:
Wow, I didn't know of that `wml.load` function.

white-haired-uncle:
Picked that up from a forum post on how to run modified WML actions without restarting scenario.  Seemed like it'd be a handy way to load rarely used code.

Issue #751, opened 2024-04-08
enemy (at least) wrath not wrathing very wrathfully


white-haired-uncle:
I noticed recently that lesser lethargy doesn't seem to do much.  The wrath is getting set, but it doesn't seem to do anything.  I know wrath on side 1 was working in the past, I haven't checked recently.

In the attached, BZB has -2 wrath_intensity.  Serves him right.  Attack him with the dwarf just to his north and you'll see the attack GUI says he'll do 10 damage, and he does.

Reload, and attack BZB with Argan's great underwear (advance him and take crossbow marksman first?).  In my case, BZB's wrath intensity went to -5.  Attack him with the dwarf just to his north and you'll see the attack GUI says he'll do 10 damage, and he does.

[LotI1-Siege Turn 10.gz](https://github.com/Dugy/Legend_of_the_Invincibles/files/14910297/LotI1-Siege.Turn.10.gz)

Dugy:
Lethargy and wrath don't work on units controlled by AI. This is not intended.

white-haired-uncle:
Not intended as in it's not working correctly, or it's not intended that wrath/lethargy work for AI units (in which case, why have a lesser lethargy special available to the player)?

Dugy:
Not intended as it's not working correctly. Human controlled units are given a hidden weapon special called latent_wrath which causes the wrath/lethargy variable to have an effect. It's missing on other units, probably as an oversight while moving that code into lua.

The latent_wrath weapon special should be changed to an ability and added to every unit without it by the code that modifies the wrath variable.

white-haired-uncle:
I now see how the latent_wrath works and why AI units don't.

What I don't get is how you would change that from a weapon special to an ability.  Don't you need the [damage] weapon special to adjust the damage (before slow/resistance/etc)?

Dugy:
If a weapon special appears in `[abilities]`, it behaves as an ability that grants the weapon special. It's a relatively new addition.

white-haired-uncle:
Cool.  I'm starting to step through this now, I've moved it from attacks to abilities and I'm starting in on the various levels of events.  I'm really wondering if there is (or will be) any reason to have unit.variables.wrath_intensity now, since it's (probably) just going to match unit.abilities.damage.add.

Dugy:
> I'm really wondering if there is (or will be) any reason to have unit.variables.wrath_intensity now, since it's (probably) just going to match unit.abilities.damage.add.

[...]

PR #792, opened 2024-05-10
fix wrath for enemies, move to abilities

white-haired-uncle:
I'm not convinced the changes to update_stats are even needed anymore, since the ability special is created if it doesn't exist when it is needed.

Creating the ability on demand also fixed the issue with enemies not being effected by wrath, at least when attacked with lesser lethargy.

I don't see any reason all of the wml setting unit.variables.wrath* can't be moved to [set_wrath_intensity], assuming we can't just get rid of the variables stuff outright.  If not, I should look at how the unit is passed so that we don't need to "unstore" twice.

Fueled by death appears to still only look at positive wrath.

Tested with whirlwind and greater underdraw (lesser lethargy).

See #751

### DIFF
--- a/lua/main.lua
+++ b/lua/main.lua
@@ -1062,18 +1062,45 @@ function loti.util.list_attacks(unit)
 end
 
 function wesnoth.wml_actions.set_wrath_intensity(cfg)
+	local debug = cfg.debug or false
 	local unit_id = cfg.id or wml.error("[set_wrath_intensity]: missing required id=")
 	local unit = wesnoth.units.get(unit_id).__cfg
 	local vars = wml.get_child(unit, "variables")
 	local abilities = wml.get_child(unit, "abilities")
+
 	local wrath_intensity = 0
-	if vars.wrath == true then wrath_intensity = vars.wrath_intensity end
+	if vars.wrath == true then
+		wrath_intensity = vars.wrath_intensity or 0 -- Do these still need to be separate?
+		if debug then wesnoth.interface.add_chat_message(string.format("wrath_intensity was %d",wrath_intensity)) end
+		if cfg.set ~= nil then wrath_intensity = cfg.set end
+		if cfg.div ~= nil then
+			if math.abs(wrath_intensity) <= 1 then
+				wrath_intensity = 0
+			else
+				wrath_intensity = wrath_intensity / cfg.div
+				if wrath_intensity > 0 then
+					wrath_intensity = math.floor(wrath_intensity)
+				else
+					wrath_intensity = math.ceil(wrath_intensity)
+				end
+			end
+		end
+		if cfg.add ~= nil then wrath_intensity = wrath_intensity + cfg.add end
+		if cfg.sub ~= nil then wrath_intensity = wrath_intensity - cfg.sub end
+		vars.wrath_intensity = wrath_intensity
+		if debug then wesnoth.interface.add_chat_message(string.format("\twrath_intensity is %d",wrath_intensity)) end
+		if vars.wrath_intensity == 0 then
+			vars.wrath_intensity = nil
+			vars.wrath = nil
+		end
+	end
+
 	local latent_wrath_special = wml.get_child(abilities, "damage", "latent_wrath")
 	if latent_wrath_special == nil then
-		--wesnoth.interface.add_chat_message("Didn't find latent_wrath_special, creating")
+		--wesnoth.interface.add_chat_message(string.format("Didn't find latent_wrath_special, creating with add = %d",wrath_intensity))
 		table.insert(abilities, { "damage", { id = "latent_wrath", apply_to = "self", add = wrath_intensity }})
 	else
-		--wesnoth.interface.add_chat_message(string.format("Found latent_wrath_special, adding latent_wrath_special.add = %d",wrath_intensity))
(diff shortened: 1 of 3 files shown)


white-haired-uncle:
I'm not seeing any backward compatability issue when starting from beginning of scenario.  

If you loaded an old save with active wrath as weapon special I suspect it would remain until the end of the scenario (or longer for chapter5 and possibly spinoffs).  update_stats might clear that up.

Dugy:
No, there would not be such a compatibility issue because the events providing the wrath decay behaviour would remain imprinted in the scenario code.

white-haired-uncle:
Replaced all of the WML adjustments to unit.variables.wrath_intensity with add=/div= options to [set_wrath_intensity].

I'm still not convinced the unit.variables are needed, and I suspect setting unit.variables.wrath could be moved to  [set_wrath_intensity] (removing the need for an unstore_unit), but it's probably not necessary to optimize everything at this point, we've moved the specials to abilities, simplified the WML quite a bit, and fixed a couple things in the process.

fueled and at least one other that weren't considering negative wrath values should be fixed as a side effect.

tested with whirlwind and lesser lethargy.

Dugy:
I really like your design.

Just a few implementation details:

1. The intensity is now set right in the ability or is it the one in the unit's `[variables]`? You seem to have the information duplicated in the two.
2. If it decays to zero, the ability may be removed.

The reason why I think it's better to keep the intensity in the unit's `[variables]` is that it would allow moving the declaration of wrath outside of lua, use a macro defining that ability as an argument to the WML tag, and reuse the code to create other abilities that can accumulate and decay. But I don't insist you do it this way, it's also possible to do it entirely in lua and use a callback to select the property that changes.

white-haired-uncle:
> I really like your design.
> 
> Just a few implementation details:
> 
>     1. The intensity is now set right in the ability or is it the one in the unit's `[variables]`? You seem to have the information duplicated in the two.
> 

It is duplicated.  The way it was done before is the value in the variables was adjusted based on the previous value, and then the result was copied to [damage].  I just copied that (more or less) so it's effectively the same as before, but with one copy of [damage] instead of one per attack.  I don't see a point to this, which is why I was asking earlier if it needed to remain that way.
variables.wrath is the only thing that really "needs" to be in variables to serve as a filter for decay (and I'm not even sure that's true, I just didn't push the issue).

>     2. If it decays to zero, the ability may be removed.

I hadn't thought of that, and I didn't initially see a reason, but I do now.  The existence of the ability could serve as a semaphore, eliminating the need for variables.wrath.  I'm not sure, but it might even be more efficient to filter on an ability vs a unit.variable?

Then again, IMO the variables stuff was redundant before and it really isn't hurting anything so I can live with it.
 
> 
> The reason why I think it's better to keep the intensity in the unit's `[variables]` is that it would allow moving the declaration of wrath outside of lua

I don't see why having it only in [abilities][damage] would preclude accessing it in WML, though it's a little more work.

> , use a macro defining that ability as an argument to the WML tag, and reuse the code to create other abilities that can accumulate and decay. But I don't insist you do it this way, it's also possible to do it entirely in lua and use a callback to select the property that changes.

You lost me here.  Though as far as reusing the code, I think it would be pretty simple to change most of what is now [set_wrath_intensity] into something generic that took "wrath" as one possible input, if there was a second similar parameter that would grow/shrink/decay in the same way.

Dugy:
> It is duplicated. The way it was done before is the value in the variables was adjusted based on the previous value, and then the result was copied to [damage]. I just copied that (more or less) so it's effectively the same as before, but with one copy of [damage] instead of one per attack. I don't see a point to this, which is why I was asking earlier if it needed to remain that way.

Keeping it in `[variables]` is superfluous now. It's not even needed for backwards compatibility because the WML code in events is saved with the mid-scenario save file.

>  The existence of the ability could serve as a semaphore, eliminating the need for variables.wrath. I'm not sure, but it might even be more efficient to filter on an ability vs a unit.variable?

Yes. I think the performance difference would be marginal, but it would avoid some clutter.

> Then again, IMO the variables stuff was redundant before and it really isn't hurting anything so I can live with it.

Originally, every unit had a latent_wrath weapon special on every attack, and the variable was just adjusting how they affected damage. There are better ways of doing that now.

> I don't see why having it only in [abilities][damage] would preclude accessing it in WML, though it's a little more work.

My idea was doing something like this:
```[set_stackable_ability]
  add=1
  id=$unit.id
  {ABILITY_WRATH}
```[/set_stackable_ability]

white-haired-uncle:
I've removed [variables], everything is now under [abilities][damage] which is removed when not needed and serves as a filter for the decay event.

[set_stackable_ability] sounds like an interesting exercise, but AFAIK there is no other stackable ability right now and I've got other things to work on.

Dugy:
Does it work with negative numbers (lethargy)?

white-haired-uncle:
Now that you mention it, that's all I tested (whirlwind and lesser lethargy).  I went back and tested wrath, and a combination (wrath+whirlwind).  They work too.

Dugy:
All right. I suppose that `add=` accepts negative numbers and will then work as subtract.

white-haired-uncle:
It should, but there's also a sub=.

PR #798, opened 2024-05-14
lilith appears, statues awaken

### LINKED ISSUE #796: Nightmare Cellar
I just wanted to change status=not_living, but that led me to Nightmare Cellar where I'm seeing a number of potential problems.  After fixing Lilith's side so that she appears like she's supposed to, I'm seeing these:

```
20240514 00:13:30 error engine: Formula in WML string cannot be evaluated due to Expected another token
	--> "
20240514 00:13:30 error engine: Formula in WML string cannot be evaluated due to Illegal unary operator: '*'
	--> "
```

which I suspect are caused by quests.lilith_deaths not existing.  I debugged my way to this scenario, but looking at the code I think this will happen the first time you meet her (at least) no matter how you get there.

https://github.com/Dugy/Legend_of_the_Invincibles/blob/8dae417d6df6f2e0a3496872008e8d3af22574ec/utils/chapter9_utils.cfg#L571-L579

I actually first thought that error might be coming from this:

https://github.com/Dugy/Legend_of_the_Invincibles/blob/8dae417d6df6f2e0a3496872008e8d3af22574ec/utils/chapter9_utils.cfg#L516

I assume that's supposed to be met_lilith ?  It's not the cause of my current error, but I still think it's wrong.

white-haired-uncle:
Nightmare Cellar:

Fixed Lilith's side so she appears

Moved statues slightly to provide a little separation and move unit out of wall.

Changed activation to use radius and proper locations.

The gender_X_Y variables can't be cleaned up properly, so I hid them under quests.  I have this thing about orphaned variables cluttering up the inspect gui.

Closes #796

### DIFF
--- a/scenarios9/30_Nightmare_Cellar.cfg
+++ b/scenarios9/30_Nightmare_Cellar.cfg
@@ -74,7 +74,7 @@
         {RARE_ITEM_TURN 28 8 -999}
         [if]
             [variable]
-                name=quests.entered_sc30
+                name=quests.sc30.entered
                 equals=yes
             [/variable]
             [else]
@@ -86,13 +86,13 @@
                     speaker=Lethalia
                     message= _ "Collectors' items, contemporary art... A stroll through her museum, that is all we needed. We better go back and pay a courtesy visit to the artist herself in Pandemonium..."
                 [/message]
-                {VARIABLE quests.entered_sc30 yes}
+                {VARIABLE quests.sc30.entered yes}
             [/else]
         [/if]
     [/event]
 
     [event]
-        name=before fist activation of statues
+        name=before first activation of statues
         [message]
             speaker=Efraim
             message= _ "These statues look somewhat... menacing."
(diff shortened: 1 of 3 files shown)


Dugy (approved):

Dugy:
> The gender_X_Y variables can't be cleaned up properly, so I hid them under quests. I have this thing about orphaned variables cluttering up the inspect gui.

I know that problem, but I don't have a good memory for these things. I think that one of the spinoff campaigns has a macro for defining a variable and simultaneously placing an event that would clear it at scenario end.

white-haired-uncle:
The problem here is that the variable can't be removed when you leave this scenario if the unit has not been activated and the player may return to the scenario.  It needs to exist when they return.  And since there's no way to know if they will return, we can't clear it.  So I just hid it.

I also like to shove stuff like that under in one table so I can see it all in one place (and save to clipboard) in the inspect gui.  I find inspect to be quite useful, but it's kind of quirky.

PR #801, opened 2024-05-17
Wrath

white-haired-uncle:
Fix error where ability only removed with debug option on.

Handle changing wrath intensity on units on recall (should not happen now that above bug is fixed, however)


white-haired-uncle:
closes #751

PR #806, opened 2024-05-23
correct corruption4 regeneration text

white-haired-uncle:
Also, note that in the main GUI if you mouse over her level you see "Oops..." where advancements are supposed to be.

### DIFF
--- a/units/Vritra.cfg
+++ b/units/Vritra.cfg
@@ -710,7 +710,7 @@ A unit afflicted by this ability may lose up to 16 HP per turn, except for poiso
                 id=regenerates
                 name= _ "regenerates"
                 female_name= _ "female^regenerates"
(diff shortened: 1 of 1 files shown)


Dugy:
> Also, note that in the main GUI if you mouse over her level you see "Oops..." where advancements are supposed to be.

That is a placeholder text that shows if something goes wrong. It can be caused by all kinds of problems. Do you have a save file where it can be seen?

white-haired-uncle:
[LotI2-Jungle Hell-Auto-Save64.gz](https://github.com/Dugy/Legend_of_the_Invincibles/files/15421614/LotI2-Jungle.Hell-Auto-Save64.gz)

Dugy:
I fixed that.

PR #807, opened 2024-05-24
exterminator steals, not shadowwalker

white-haired-uncle:
I didn't know the thief line also had steals by default.  Be nice if they had the ability so we could see it.

Anyway, attacker hits had exterminator, but defender hits had shadowwalker.


Dugy:
Yes, that's how it should correctly be.

PR #809, opened 2024-05-25
dynamically load most bzb events

white-haired-uncle:
Of the five bzb events, I made three load on demand, the others were really short.  I don't like inconsistency, but it seemed to make sense for now.

Loading WML like this means no pre-processor (macros).  It's not hard to convert using the save file (or perhaps wesnoth -p), but you have to be careful not to overlook anything, and it can be a little tedious to replace the variable substitution.  I suspect anything that relies on difficulty level ({QUANTITY}, {GOLD} etc) is going to be annoying if done by hand but I don't see any reason it couldn't be automated.

The savings here are not impressive, about 8KB compressed, which is around 1.3% in Gates of Hell, and 2.5% in Toxic Sun.    Probably not worth doing on its own, but going through the process may lead to more general methods and procedures (or lead to the determination that it's not worth doing).

When I did the help menu stuff, I moved those into help/.  In this case, I put the "events" in dynamic/beelzebub.  I'm not really sure what the layout should look like, I'd probably have to do a few more to get an idea of what makes sense.  I do kind of like taking things out of a monolithic file like utils.cfg and breaking them into a hierarchy of smaller files, iff said hierarchy makes sense.  For example, for bzb there are currently two files that are basically message/option blocks, and one I'd call an event, so would they go in dynamic/events/beelzebub and dynamic/menus/beelzebub, or dynamic/events/ and dynamic/menus, or dynamic/beelzebub/events|menus or just dynamic/beelzebub, or something completely different (perhaps under utils/)?  I have no idea at this point, except that if there's a dynamic/events/ there shouldn't also be a dynamic/beelzebub/events -- that would bug me.


Dugy:
I don't think that creating a toplevel folder named `dynamic` just because of this is a good way of doing things. Can you just put it into `utils` (into a specific subfolder or not)?

white-haired-uncle:
Sure.  I should probably move help/ as well.

ref #787

Dugy:
Yeah, makes sense.

Dugy (approved):

PR #810, opened 2024-05-25
Teleporting to Lethalia in Across the Barren Land

konecnyjakub:
In Across the Barren Land, there is a message and a note in objectives that you can teleport your units to your leader via a right click menu item. But it  is always added to Efraim even if you have chosen to play Lethalia so in that case you do not have that option. Here is a fix.


Dugy:
Thanks. I think I didn't feel a need to do teleport them while I was playtesting.

PR #814, opened 2024-05-25
spelling

white-haired-uncle:


white-haired-uncle:
I was actually looking to see if a dark shade casts a shadow (could you use a pair of them to boost each other every turn up to the limit).

Seems to me that perhaps some units shouldn't, like ghosts (maybe they would, since they seem to be wearing robes).  Probably  anything that darkens.  Almost certainly anything that is currently invisible.

Dugy:
> I was actually looking to see if a dark shade casts a shadow (could you use a pair of them to boost each other every turn up to the limit).

Dark Shade is made invisible only by units that have no ability to hide. So, for example, if a Dark Shade sits around an Elvish Avenger and no other unit, it is visible even if the Elvish Avenger isn't actually hidden.

I didn't want two Dark Shades to be able to help each other stay invisible. I originally wanted to make them hidden when next to a unit that isn't hidden, but when a Dark Shade stood one next to the other, both hid because they were next to another visible unit, but then both became visible because they had no nearby visible units, then hid again... and the game crashed, probably with a stack overflow.

PR #815, opened 2024-05-26
blizzart does not move units, unfortunately

white-haired-uncle:
I know I fixed this as some point.  Maybe it was in WML?


Dugy:
I might have been thinking of that at some point, but I decided not to do that. Probably because the implementation would be overcomplicated.

PR #816, opened 2024-05-26
unplaguable > unplagueable

white-haired-uncle:
If this section encounters a status it doesn't understand (like unplaguable), line simply doesn't change.  So for Soul Eclipise, it showed "Immune to Drain" twice.  Maybe you want to add an else line=<I don't understand $part> or whatnot?


Dugy:
Statuses that are not understood should be ignored. There are some statuses in the core game that aren't shown.

white-haired-uncle:
That would work too, but currently they're treated as whatever the last known one was.

Dugy:
Then it's a bug.

Dugy:
I wasn't able to replicate that problem. Unknown statuses don't seem to show for me.

white-haired-uncle:
Did you first have a known status?  You need it so line gets set to something.  For example, edit Soul Eclipse:

add=unpoisonable,undrainable,untenable

Should show

Immune to Poison
Immune to Drain
Immune to Drain

because when it gets to untenable and doesn't find a match then line remains "Immune to Drain" and gets printed again.

Dugy:
I thought you had the status icons in mind, not the item descriptions. I fixed that now.

PR #823, opened 2024-06-23
fix faster wail

white-haired-uncle:
I'm not sure this attack is working correctly.  From what I gather from the GUI, it should only be active if the unit ate a WC this turn, but it seems to always be active.


Dugy:
It definitely shouldn't be always active, only during the turn it ate a Walking Corpse.

white-haired-uncle:
It works correctly in the latest build.

PR #831, opened 2024-07-29
Fix the bug where advancing during enemy turn prevent from altering AMLA

rodolphecombe:
As the title states, this quick one-line fix is to allow players to alter advancements chosen for units that leveled-up during enemy turn.
AFAIK there is no other consequences


Dugy:
Thank you.

PR #838, opened 2024-09-21
Ch3 Bitter Swamp: buying the Tome of Necromancy now gives you the right item.

Chergnobarp:
I was quite surprised to get Warheart when I bought the tome. After a brief look at git blame, I have no idea how this came about.


Dugy:
This happened because I've been changing the numbers of some items a very long time ago and I might have reused some of these numbers.

PR #846, opened 2025-03-25
Critical fix of translation error causing crashes

Bafy78:
Kinda stupid lol, I thought the 3rd "%" were actual percent symbols, but Ig they aren't.

Btw I'm starting my second playthrough of the campaign ; expect translations improvements in the future!


Dugy:
I don't understand the problem, nice thing you found how to fix it.

Bafy78:
I don't understand the problem either. I was getting crashes when trying to start a new game, at the tutorial, but only with the new french translation. I narrowed it down to these lines with a binary search. Idk what was the problem and don't need to know x)

PR #853, opened 2025-07-30
Added a Beelzebub spawn point in ch9.

Chergnobarp:
I triggered Beelzebub at the end of ch8, and got shafted out of the encounter. I tried to follow the logic from Akula's place, I can relocate it easily enough if you want.


Dugy:
Yeah, that kind of correction is fine. Thanks!

PR #854, opened 2025-08-02
Made nature's power work on infernal forest terrain.

Chergnobarp:


Dugy:
I thought I set it that way on purpose, but if you think it's better that way, then okay.

Chergnobarp:
I'm happy with either behavior, it's more a UI thing for me. When the terrain shows as *Infernal Forest (Forest, Flat)*, the user expects forest powers to work there.

PR #861, opened 2025-12-18
Spanish image translation and some fixes

Sebas38760:
1. main.cfg: Separation of proper names and translatable text
2. utils.cfg: Macro separated from translatable string
3. images: Created l10n/es directory
4. images/l10n/es: Added chapter-text images translated into Spanish
5. images/l10n/es: Removed temporary placeholder file
6. Spanish translation updated


Dugy:
Seems mostly fine, but what's the problem with credits that you need fixing?

Sebas38760:
> Seems mostly fine, but what's the problem with credits that you need fixing?

There was no problem with the credits themselves.
I only separated the texts (proofreading), (dialogues), and (mechanics) so they could be translated while keeping the names intact.

PR #873, opened 2026-05-27
Item sorts are now translatable

rodolphecombe:
Solve the TODO about making item sorts translatable in the item menu.
Tested with French translation. 
I can update the French .po if needed.

### DIFF
--- a/lua/inventory/storage.lua
+++ b/lua/inventory/storage.lua
@@ -296,6 +296,32 @@ local function unhide_listbox(dialog)
 	dialog[listbox_id]:focus()
 end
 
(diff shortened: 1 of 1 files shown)


Dugy:
Thanks. If the French .po file was written in a way that it doesn't work with this change, feel free to update it here.

rodolphecombe:
The PR is adding new strings (the translatable strings representing the item sorts), so the .pot and all .po need to be updated.

Dugy:
All right, if you can update the .po, it would help me greatly, I don't know how is it done.

PR #874, opened 2026-05-29
Update French translation

rodolphecombe:
Follow-up of #873


Dugy:
Thank you for bringing this to the finish line.

PR #882, opened 2026-08-08
Status effects now cleared when advancing on the enemy's turn.

Chergnobarp:
Should I add petrified to the list? Units could still advance via incineration or damage aura.


Dugy:
Are you sure this clears the statuses rather than sets them all to `no`? I think a better way to do it would be to erase the unit's `[status]` array before unstoring it at some point.

Also, I think that petrify could be removed as well. The unit can be attacked until it advances.

Dugy:
I have cleared the status to the version of the unit that is going to be advanced, which would regenerate it without knowing anything that was set into the unit.

